PR #1, opened 2019-12-24
Add the model, migration and api for games

tychota:
# What

This PR adds: 
- a `games` django applications (that contains everything regarding go games)
- some game-like model, like "a selfplay" game, a "match" game
- the admin to modify /edit those data
- the api to properly access the data

# How to test

The api now have "matches" and "selfplay" section

image

image

The admin have a match section

image


config/urls.py
```diff
@@ -31,6 +31,7 @@
         ),
         # Django Admin, use {% url 'admin:index' %}
         path(settings.ADMIN_URL, admin.site.urls),
+        path(f'{settings.ADMIN_URL}/doc/', include('django.contrib.admindocs.urls')),
```
lightvector: Question: What does this do or enable?

katago_server/games/admin.py
```diff
@@ -0,0 +1,12 @@
+from django.contrib import admin
+from katago_server.games.models import SelfPlay, Match
+
+
+@admin.register(SelfPlay)
```
lightvector: Can you explain what the relationship is between the SelfPlay and SelfPlayAdmin?

katago_server/games/migrations/0001_initial.py
```diff
@@ -0,0 +1,43 @@
...
+
+class Migration(migrations.Migration):
+
+    initial = True
+
+    dependencies = [
+    ]
+
+    operations = [
+        migrations.CreateModel(
```
lightvector: The reason we need a migration already here initially is because we are "migrating" from the "empty model" to the initial model that actually has this stuff?

config/urls.py
tychota: It enable an handi page that use the comment to write documentation about model and view.

See an intro here:
https://nearsoft.com/blog/the-secret-of-django-auto-generated-admin-docs/

I used it to see what model i should remove here: https://github.com/tychota/katago-server/commit/3adaa6de03f1cfe0fa4dec61a18c933ec94e5a79

It does work in dev and should work in prod (but isn't working because a missing deps).

katago_server/games/admin.py
tychota: Selfplay is the model. It generate sql base on an orm, and create a table in db (simplistic explanation)

SelplayAdmin is an admin object. 

When saying 

@admin.register(model)

We add a new section in django admin (see discord for credentials)

tychota: We then can customize the admin itself by using the class under.

tychota: See https://djangobook.com/mdj2-django-admin/ for more details / example.

katago_server/games/migrations/0001_initial.py
tychota: All change to db schema should be a mutation. So you can migrate rollback a change.

Normally you should mutate a existing mutation. I didn't respect this here since the model is changing fast wich generate nasty migration change (eg if you generate "selfplay" model, then create "network" and add the foreign key "network_id" in "selfplay" django will ask you smartly to prpvide default in migration so old "selfplay" can have a network associated. Which is smart when you have a lot of data but more complexe as the start than resetting db and redoibg a clean init migration).

PR #2, opened 2019-12-24
Add training app

tychota:


katago_server/trainings/models.py
```diff
@@ -0,0 +1,37 @@
...
+    nb_requested_matches = IntegerField()
+    nb_won_by_reference_matches = IntegerField()
+    nb_won_by_testes_matches = IntegerField()
+    nb_draw_matches = IntegerField()
+    nb_no_result_matches = IntegerField()
+
+    created = DateTimeField(auto_now_add=True)
+    modified = DateTimeField(auto_now=True)
+
+    elo_difference = IntegerField()
```
lightvector: If we are going to report Elo difference it should probably have absolutely nothing to do with gating. Otherwise, it's very, very hard to avoid bias. One would want an entirely separate set of matches such as CloudyGo does. Although, I'm not sure we need to dive into that complexity just right away.

katago_server/games/models.py
```diff
@@ -64,25 +66,37 @@ class Game(Model):
...
 class SelfPlay(Game):
+    network = ForeignKey(Network, on_delete=PROTECT)
+
     unpacked_training_file = FileField()
     packed_training_file = FileField()
 
+    game_extra_params = JSONField()
+
+
+class ForkedSelfPlay(Game):
```
lightvector: As I mentioned in other comments - I think we can simplify and eliminate entirely the notion of a "forked" game so long as it is understood that "game" might start in a position other than the empty position, and might start at a turn number other than 0. If we want to distinguish forked games from other games via plots, that can be done by a simple metadata, exactly the same way one would distinguish games with one komi from games with a different komi, or games with one rules with games with a different rules.

katago_server/trainings/models.py
tychota: I wil remove the gating alltogether as discuss.

katago_server/games/models.py
tychota: Agreed, and i did remove forked game in a later commit.

PR #9, opened 2020-03-29
Models review

lightvector:
I completed my review of all the models in the server. I left a bunch of questions, which I committed in this pull request with a "QUESTION" tag on the appropriate line. For example, a shell command line `grep QUESTION . -r --include=*.py` from the root of this repo will find all instances, and you can also see it on the Github diff here.

Additionally, I changed a small number of defaults and I added a `game_hash` field  to the game which is intended to correspond to KataGo's `gameHash` here:
https://github.com/lightvector/KataGo/blob/master/cpp/dataio/trainingwrite.h#L51
I tried makemigrations but it did not seem to generate any migration file? So is this already correct?

This pull request should probably not be accepted as-is with the various QUESTIONs but I made it a pull request for better visibility and commenting.

Lastly, I rebased and force-pushed the [distributed](https://github.com/lightvector/KataGo/tree/distributed) branch of KataGo. I incorporated some of the cleanups and your cmake changes directly into master already, and the remaining changes I cleaned up and rebased them on to the new master as a clean new stack of changes. Please use the new tip for any new implementation or testing.

I think the types in the server look good, except for the questions I had. Great! What's next?


katago_server/trainings/models/network.py
```diff
@@ -11,10 +12,35 @@
 from katago_server.trainings.managers.network_pd_manager import NetworkPdManager
 from katago_server.trainings.managers.network_queryset import NetworkQuerySet
 
-
+# QUESTION (lightvector): Why does the server need to be uploading networks anywhere? Shouldn't it be the
+# case that simply networks are uploaded to the appropriate filestorage solution directly by training and
+# appropriately the server is notified?
```
tychota, line 17: Nice question. In my mind, the  training will also do a multipart file data api call, same as npz files

katago_server/trainings/models/network.py
```diff
@@ -11,10 +12,35 @@
...
 def upload_network_to(instance, _filename):
     return os.path.join("networks", f"{instance.uuid}.gz")
 
+# QUESTION (lightvector): I added these functions, are they useful for you, or do we want to do this a different way?
+# They will parse a string like "g170-b40c256x2-s2436610304-d726429752" and return
+# the blocks and channels and other things.
+def parse_katago_training_model_name(name):
```
tychota, line 24: Perfect

katago_server/trainings/models/network.py
```diff
@@ -11,10 +12,35 @@
...
+        matched = re.fullmatch(r"^b(\d+)c(\d+).*$",pieces[1])
+        if matched is None:
+            return {}
+        parsed["nb_blocks"] = int(matched.group(1))
+        parsed["nb_channels"] = int(matched.group(2))
+        parsed["nb_trained_samples"] = int(pieces[2][1:])
+        parsed["nb_data_samples"] = int(pieces[3][1:])
+        return parsed
+    except ValueError:
+        return None
```
tychota, line 42: Shouldn't it be return {}

katago_server/trainings/models/network.py
```diff
@@ -41,11 +67,16 @@ class Meta:
         verbose_name = _("Network")
         verbose_name_plural = _("Networks")
 
+    # QUESTION (lightvector): Do any changes need to be made such that this can be a different identifier supplied by the training process?
```
tychota, line 70: If it doesn't match an uuid, I should only convert it as Charfield

katago_server/trainings/models/network.py
```diff
@@ -41,11 +67,16 @@ class Meta:
...
     uuid = UUIDField(_("unique identifier"), default=uuid.uuid4, db_index=True)
     created_at = DateTimeField(_("creation date"), auto_now_add=True)
+    # QUESTION (lightvector): What is parent_network for?
     parent_network = ForeignKey("self", null=True, blank=True, related_name="variants", on_delete=PROTECT)
```
tychota, line 74: We discussed that a while ago. We need parent network for bayesian prior of bayesian elo. 

Right now the parent is always network N-1 but in case of official and community run we have

image

katago_server/trainings/models/network.py
```diff
@@ -60,6 +91,7 @@ class Meta:
     )  # used to select best unsure network for ranking games (matches)
 
     def __str__(self):
+        # QUESTION (lightvector): self.id is a mismatch with uuid? Is there also an "id" field too?
```
tychota, line 94: There is also an id, which is by default a primary key, integer autoincrementing.

katago_server/trainings/models/network.py
```diff
@@ -41,11 +67,16 @@ class Meta:
...
     created_at = DateTimeField(_("creation date"), auto_now_add=True)
+    # QUESTION (lightvector): What is parent_network for?
     parent_network = ForeignKey("self", null=True, blank=True, related_name="variants", on_delete=PROTECT)
     # Some description of the network itself
+    # QUESTION (lightvector): Theoretically one day we could have an architecture that is different than a resnet, do we want to make blocks optional?
     nb_blocks = IntegerField(_("number of blocks in network"))
+    # QUESTION (lightvector): Not just theoretically, but actually it is experimentally possible for one to try networks with variable numbers of
+    # channels in different layers of different blocks, like "wide" resnets and such. Do we want to make channels optional?
     nb_channels = IntegerField(_("number of channels in network"))
     model_architecture_details = JSONField(_("network architecture schema"), null=True, blank=True, default=dict)
```
tychota, line 81: I think we should maybe just replace "nb_bloc" and "nb_channel" by a string, eg:

"b20c192"
"b20c[168-256] (wide)"

We loose a bit of UX (user cannot order network by size for instance) but it is simpler and a bit more flexible.

katago_server/distributed_efforts/models/configuration/dynamic_distributed_task.py
```diff
@@ -15,6 +15,9 @@ class Meta:
     katago_config = TextField(
         _("katago config"), help_text=_("The configuration file for katago to be given to client for selfplay"), default="FILL ME"
     )
+    # QUESTION (lightvector): I assume "probability_predefined_training_game" is supposed to be the games that start from specified SGF positions?
```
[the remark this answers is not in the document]
lightvector: Okay, I thought it was weird that we would have ranking games start in predefined positions. But I guess in theory it could be used to force starting with different opening moves or something.

katago_server/distributed_efforts/models/distributed_task/abstract_distributed_task.py
```diff
@@ -12,6 +12,8 @@
 from katago_server.trainings.models import Network
 from katago_server.users.models import User
 
+# QUESTION (lightvector): What does this mean? Does this mean if the game doesn't complete within an hour, it is assumed to be forgotten?
```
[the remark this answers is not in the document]
lightvector: What happens if the client does actually reply back after it expires?

katago_server/distributed_efforts/models/distributed_task/ranking_estimation_game_distributed_task.py
```diff
@@ -25,7 +25,7 @@ def create_with_random_color(cls, reference_network, opponent_network):
         :param opponent_network: Second network
         :return:
         """
-        if random.random() < 0.5:  # TODO: should that 0.5 be configurable ?
+        if random.random() < 0.5:
```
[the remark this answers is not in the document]
lightvector: Yes. I see no reason why we would want to change this.

katago_server/trainings/models/network.py
lightvector: Will fix thanks.

katago_server/trainings/models/network.py
lightvector: I'm not sure what this means, but I'll let you handle this. So if the training machine is going to upload a network, presumably the API will also have the training machine at runtime tell the server through the API call what the model's name is, which could be used as the key, and additionally at that time the server would also call `parse_katago_training_model_name` on it to extract out possible the other details, if present.

katago_server/trainings/models/network.py
lightvector: Ah, right. Okay. :)

katago_server/trainings/models/network.py
lightvector: Nevermind, I just went ahead and tried this myself. Still not sure where `parse_katago_training_model_name` should be called, but maybe even easier is if the training machine also just provides these values separately?

lightvector:
Pushed two more commits and removed the QUESTION comments I added. If you think this looks good, please merge, otherwise let me know what I should fix. Thanks!

katago_server/distributed_efforts/models/configuration/dynamic_distributed_task.py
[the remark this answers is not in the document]
tychota: Exactly

katago_server/distributed_efforts/models/distributed_task/abstract_distributed_task.py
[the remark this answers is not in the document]
tychota: I guess it will be discarded.

tychota:
Sure, will merge it today.

PR #12, opened 2020-05-11
More cleanups

lightvector:
* Adds NetworkSerializerForClient.
* Updates pairing logic in cases of having no networks or very few networks.
* Adds handling for no active run in more places.


lightvector:
Also allows blank nets for the random net placeholder.

katago_server/runs/models/run.py
```diff
@@ -46,7 +46,9 @@ class RunStatus(TextChoices):
     id = AutoField(primary_key=True)
     created_at = DateTimeField(_("creation date"), auto_now_add=True, db_index=True)
     status = CharField(_("run status"), max_length=15, choices=RunStatus.choices, db_index=True, default=RunStatus.INACTIVE,)
-    # TODO enforce that name is UNIQUE
+    # TODO is it possible to make this the primary key?
```
tychota, line 49: The proper way to do this is:

- in the viewset

```diff
 class RunViewSet(viewsets.ModelViewSet):
     """
     API endpoint that allows runs to be viewed or edited.
     """
     queryset = Run.objects.all()
     permission_classes = [IsAdminUser | ReadOnly]
     serializer_class = RunSerializer
     filterset_fields = ["status"]
+    lookup_filed = 'name'
```

and not to break the automatically generated url, in each serializers

```diff
 class RunSerializerForClient(HyperlinkedModelSerializer):
     """
     Serializer exposing only the fields of a run that a self-play client needs on startup.
     """
 
     class Meta:
         model = Run
         fields = [
             "id",
             "url",
             "name",
             "data_board_len",
             "inputs_version",
             "max_search_threads_allowed",
         ]
+       extra_kwargs = {
+            'url': {'lookup_field': 'licensenumber'}
+        }
```

Last important point (but already done here) is to have `db_index=True` in the model so the query remains efficient.

katago_server/runs/serializers/run.py
```diff
@@ -21,3 +21,20 @@ class Meta:
...
+
+
+class RunSerializerForClient(HyperlinkedModelSerializer):
+    """
+    Serializer exposing only the fields of a run that a self-play client needs on startup.
+    """
+
+    class Meta:
+        model = Run
+        fields = [
```
tychota, line 33: We should try to unify the code.
Sometimes we use tupple (), sometimes list.

I don't know what is best and why but we should find out and stick to it

katago_server/runs/viewsets/run.py
```diff
@@ -31,3 +31,14 @@ def current(self, request):
             return Response({'error': 'No active run.'}, status=status.HTTP_404_NOT_FOUND)
         self.kwargs["pk"] = current_run.pk
         return self.retrieve(request)
+
```
tychota, line 34: Maybe you can remove the current as it is no longer needed.

katago_server/trainings/models/network.py
```diff
@@ -52,6 +52,9 @@ class Meta:
         ordering = ["-created_at"]
 
     id = BigAutoField(primary_key=True)
+    # TODO is it possible to make this the primary key?
+    # Having a url like https://katago.tycoach.me/api/networks/2/ is not very useful if the intended name for users
+    # to recognize the network is not "2", but rather this name.
```
tychota, line 57: Same as before, look up for "lookup_field" :)

Also is network name really unique ? I think it is only unique inside a run so we should remove unique true and add 
https://stackoverflow.com/a/56024282

katago_server/trainings/models/network.py
```diff
@@ -75,7 +78,7 @@ class Meta:
         storage=network_data_storage,
         max_length=200,
         null=False,
-        blank=False,
+        blank=True,
```
tychota, line 81: ? why change this ?

tychota: A yes for the first net.
Then also change the `null = False` constraint.

katago_server/distributed_efforts/viewsets/distributed_task.py
```diff
@@ -19,28 +20,37 @@ class DistributedTaskViewSet(viewsets.ViewSet):
     # noinspection PyMethodMayBeStatic
     def create(self, request):
         current_run = Run.objects.select_current()
+        if current_run is None:
+            return Response({'error': 'No active run.'}, status=status.HTTP_404_NOT_FOUND)
```
tychota, line 24: Well done

katago_server/trainings/tasks/update_bayesian_rating.py
```diff
@@ -13,6 +13,8 @@ def update_bayesian_rating():
     :return:
     """
     current_run = Run.objects.select_current()
+    if current_run is None:
+        return
```
tychota, line 17: We should throw a custom error so sentry (the bug catching platform) is notified.

katago_server/trainings/models/network.py
```diff
@@ -118,4 +121,8 @@ def save(self, *args, **kwargs):
             # TODO: We should let the api to create a model throw if there is no parent mode
             if not self.parent_network:
                 self.parent_network = Network.objects.last()
+        # Allow blank file only if random
+        if (not self.is_random) and (self.model_file is None or len(self.model_file) <= 0):
```
tychota, line 125: I prefer extracting the condition and removing the comment:

```python
is_initial_random_network = (not self.is_random) and (self.model_file is None or len(self.model_file) <= 0)
if not is_initial_random_network:
  ...
```

katago_server/runs/viewsets/run.py
```diff
@@ -31,3 +31,14 @@ def current(self, request):
...
+    @action(detail=False, methods=["GET"])
+    def current_for_client(self, request):
+        """
+        API endpoint that gives only the fields of a run that self-play clients need.
+        :return:
+        """
+        current_run = Run.objects.select_current()
+        if current_run is None:
+            return Response({'error': 'No active run.'}, status=status.HTTP_404_NOT_FOUND)
+        return Response(RunSerializerForClient(current_run,context={'request': request}).data)
```
tychota, line 44: Extract `run_serilaizer = RunSerializerForClient(current_run,context={'request': request})` for better readability.

katago_server/distributed_efforts/services/rating_network_pairer.py
```diff
@@ -68,10 +75,23 @@ def _choose_opponent(self, reference_network):
...
-        if len(nearby_networks) < 5:
-            nearby_networks_a = Network.objects.filter(run=self.current_run, pk__lt=reference_network.pk).all()[:20]
-            nearby_networks_b = Network.objects.filter(run=self.current_run, pk__gt=reference_network.pk).all()[:20]
-            return np.random.choice(nearby_networks_a + nearby_networks_b)
+        if len(nearby_networks) < 4:
+            nearby_weaker_networks = (
+                Network.objects.exclude(pk=reference_network.pk)
+                .filter(run=self.current_run, log_gamma__lte=ref_net_log_gamma)
+                .order_by("-log_gamma")
+                .all()[:2]
```
tychota, line 83: I thin the right logic to `find the X closest network in elo of network Y` should be:
- find X (and not X /2) networks after Y, ranked by elo excluding Y
- find X (and not X /2) networks before Y, ranked by elo excluding Y
- compute elo diff
- order by abs(elo_diff)
- select first X

That will works if all the network are above elo Y.

In this case it means that it is 4 and not 2.
And Since it start being a complex logic, I would extract it into a querySet or a manager, like current_run function for instance.

katago_server/runs/models/run.py
lightvector: It should be "name" instead of "licensenumber", right?

katago_server/trainings/models/network.py
lightvector: I read online that null = False does not work properly for FileField, and therefore setting is not a good idea? https://docs.djangoproject.com/en/dev/ref/models/fields/#null

lightvector: (setting blank=True only does work for me, I was able to create an appropriate random net)

katago_server/trainings/tasks/update_bayesian_rating.py
lightvector: Is it a bug to have no run? Or do we simply just not want to update ratings if we're in the middle? Say we deactivate a run, but it's a few days until we begin the next run.

katago_server/distributed_efforts/viewsets/distributed_task.py
lightvector: thanks

katago_server/distributed_efforts/services/rating_network_pairer.py
lightvector: I wasn't attempting to find precisely the X closest, just the nearest two on each side I think is fine anyways. I think this is just a corner case where doing anything reasonable is okay, it won't be very important. Except for maybe a random network or the first network after random, it will never be the case networks are more than 1200 Elo apart.

katago_server/trainings/models/network.py
lightvector: Network name being unique even across runs is probably a good practice, so I'm happy to have that restriction. If we ever had multiple runs wouldn't be happy with an overlapping name for user-friendliness because then when the users run download and use it, it would be confusing to talk about the model or remember which run it came from.

katago_server/runs/models/run.py
tychota: yes, copy pasta from the doc :/

katago_server/trainings/models/network.py
tychota: But name for me is the part: `s4384473088-d968438914`.
Along with run name: `g170e` and the network info `b20c256x2`  it makes the full filename: `g170e-b20c256x2-s4384473088-d968438914.zip`

katago_server/trainings/models/network.py
tychota: Interesting: I always did the distinction. "" means a string that can be empty but null is stronger and assert the willingness to put no data, not just an empty string.

I must do some research.

katago_server/trainings/tasks/update_bayesian_rating.py
tychota: The proper way would be that `update_bayesian_rating` takes some params, for instance the run name.

Thus we can have two run in parallel and thus two task for update each run.

Line 16 would become:

```python
def update_bayesian_rating(run_name):
    run = Run.objects.get(name=run_ame)  # Will throw Run.DoesNotExist exceeption
```

katago_server/trainings/models/network.py
lightvector: When I write the network upload code, name will be like "g170e-b20c256x2-s4384473088-d968438914". Yes this is slightly redundant, but it means "name" can be the single universal identifier for the network. The KataGo code also calls this the "name" of the network when it is logging the network, or creating the network file (the network file internally contains its name, so that it can still be uniquely identified even if the user renames the file), and does not call "s4384473088-d968438914" the name.

katago_server/trainings/tasks/update_bayesian_rating.py
lightvector: How to make the `@celery_app.task()` understand to only call it on active runs if we add an argument to it?

katago_server/trainings/models/network.py
lightvector: ok

katago_server/runs/viewsets/run.py
lightvector: ok

katago_server/runs/viewsets/run.py
lightvector: ok

katago_server/distributed_efforts/services/rating_network_pairer.py
tychota: Fine for me.

katago_server/runs/serializers/run.py
tychota: Based on that https://stackoverflow.com/a/12363023 and https://stackoverflow.com/a/3671640 I think list is fine. We can clena later

katago_server/trainings/models/network.py
tychota: Mmh ok, I'm not strongly against but I think it is a bit smell code to have the duplicate. I even envisage to store s = "4384473088" and s = "968438914" separatly so we can index them and quickly build  elo graph over "d" (does that makes sense ?)

katago_server/trainings/tasks/update_bayesian_rating.py
tychota: You can configure explicitly on the admin (periodic task).

image

You can configure both positonal and keyword argument.

katago_server/trainings/viewsets/network.py
```diff
@@ -15,3 +15,6 @@ class NetworkViewSet(viewsets.ModelViewSet):
     queryset = Network.objects.all()
     permission_classes = [IsAdminUser | ReadOnly]
     serializer_class = NetworkSerializer
+    # TODO how to get this to work??
+    # lookup_field = "name"
```
tychota, line 19: What was the problem to get it work ?

katago_server/trainings/models/network.py
tychota: Fine for me

katago_server/trainings/viewsets/network.py
lightvector: Solved, the problem was my local server still had a bunch of junk fake data with non unique and with blank names, even though the field is now required to be unique.

PR #14, opened 2020-06-01
Add missing migrations

lightvector:
When I tried to build and update things from tip of master, django warned that there were migrations missing, so these are the migrations.


lightvector:
I think this was merged but github didn't detect it? So closing this.

PR #17, opened 2020-06-07
Fix elo

lightvector:
There were a bunch of bugs in the Elo code. This fixes them, and also fixes the broken API test to work and adds a few more.

* Draws were indeed needing to be handled both as white and as black. Everywhere in the code assumes the table "duplicates" how it counts games, by counting once for each side, but the old implementation mistakenly thought that draws did not need to be handled. But since everywhere assumes duplication, even though draws are symmetric they also need to count once each way.
* Anchor log gamma was mistakenly subtracted from uncertainty and from the network ids themselves perhaps, corrupting the ids? This is fixed.
* Fixed a missing sqrt for uncertainty computation.
* Fixed unnecessarily complicated pandas computations into just python for loops computing sums.
* "Fixed"?? bug where failing to call .order_by() on django db query caused aggregation not to happen. I still don't understand this one.
* Fixed performance bug where a constant dataframe was being computed every loop even though it didn't depend on the loop.
* Merge draws into counting as 0.5 of a win earlier, simplifying the code.


lightvector:
Also it gets rid of the NetworkBayesianRatingConfiguration singleton and just moves it into the run.

Bayes Elo computation should be covered now by fairly strict numeric tests - it should be much harder to accidentally break the computation now without breaking a test.

katago_server/distributed_efforts/tests/test_api.py
```diff
@@ -26,28 +42,68 @@ def test_get_job_anonymous(self):
         """
         An anonymous user should connect to get a new task
         """
-        # Given
```
tychota, line 29: Oh, I liked the "Given when Then" comments. It helped me structure the test. 
cf https://martinfowler.com/bliki/GivenWhenThen.html

Not strongly for keeping them but any reason to delete the existing ?

katago_server/distributed_efforts/tests/test_api.py
```diff
@@ -26,28 +42,68 @@ def test_get_job_anonymous(self):
...
+        assert response.status_code == 200
+
+    def test_get_job_authenticated_valid_git_revision2(self):
+        client = APIClient()
+        client.login(username="test", password="test")
+        response = client.post("/api/tasks/", {"git_revision":"1111222233334444555566667777888899990000"})
+        data = copy.deepcopy(response.data)
+        data["network"]["created_at"] = None # Suppress timestamp for test
+        data["run"]["id"] = None # Suppress id for test
+        assert str(data) == """{'kind': 'selfplay', 'run': {'id': None, 'url': 'http://testserver/api/runs/testrun/', 'name': 'testrun', 'data_board_len': 19, 'inputs_version': 7, 'max_search_threads_allowed': 8}, 'config': 'FILL ME', 'network': {'url': 'http://testserver/api/networks/testrun-randomnetwork/', 'run': 'http://testserver/api/runs/testrun/', 'name': 'testrun-randomnetwork', 'created_at': None, 'is_random': True, 'model_file': None, 'model_file_bytes': 0, 'model_file_sha256': '12341234abcdabcd56785678abcdabcd12341234abcdabcd56785678abcdabcd'}}"""
```
tychota, line 108: Really nice test overhaul.
:)

katago_server/distributed_efforts/viewsets/distributed_task.py
```diff
@@ -27,7 +27,7 @@ def create(self, request):
         git_revision_hash_whitelist = [s for s in git_revision_hash_whitelist.split("\n") if len(s) > 0]
         git_revision_hash_whitelist = [s.split("#")[0].strip().lower() for s in git_revision_hash_whitelist]
         git_revision_hash_whitelist = [s for s in git_revision_hash_whitelist if len(s) > 0]
-        git_revision = request.POST["git_revision"].strip().lower() if "git_revision" in request.POST else ""
+        git_revision = request.data["git_revision"].strip().lower() if "git_revision" in request.data else ""
```
tychota, line 30: As discussed, nice !

katago_server/games/managers/rating_game_pandas_manager.py
```diff
@@ -1,15 +1,23 @@
+import logging
 import pandas
 from django.db.models import Manager
 from django_pandas.io import read_frame
 
 from katago_server.games.managers.rating_game_pandas_queryset import RatingGamePandasQuerySet
 from katago_server.runs.models import Run
 
+logger = logging.getLogger(__name__)
```
tychota, line 9: Nice, I will make a pass to improve logging. In production we are blind right now.

I'm planning to use https://www.structlog.org/en/stable/getting-started.html
so in production we have strured log rather than text (it easier to index a json thata text).

katago_server/games/managers/rating_game_pandas_manager.py
```diff
@@ -1,15 +1,23 @@
...
+
 
 class RatingGamePandasManager(Manager):
     """
-    RatingGamePandasManager generate a tournament result using RatingGamePandasQuerySet
-    which is use to update the rating.
+    RatingGamePandasManager generates a tournament result using RatingGamePandasQuerySet
+    which is use to update the Elo ratings.
+
+    Every game played is present BOTH ways. If player 1 plays player 2, then this game
```
tychota, line 17: It is clear for me now. Thanks for the comment.
<3

katago_server/games/managers/rating_game_pandas_queryset.py
```diff
@@ -15,6 +15,7 @@ def get_total_games_count_as_white(self, run: Run):
         return (
             self.filter(run=run)
             .values(reference_network=F("white_network__pk"), opponent_network=F("black_network__pk"),)
+            .order_by()
```
tychota, line 18: That's a mystery. 
👻

katago_server/distributed_efforts/tests/test_api.py
lightvector: Hehe, interesting. They seemed unnecessary to me, sort of like
```
# Initialize x
x = 0
# Add one to x
x += 1
# Assert that x is positive
assert x > 0
```
Because as far as I can tell, "then" simply meant "here is where we begin to assert things", but you can see in the code itself that it's asserting things. And for there seemed to be not a clear distinction between "given" and "when" in our situation - for example why is `client.login(username="test", password="test")` part of "given" but not part of "when"? If "when" is always simply the last command before we begin asserting things or whose return value we are asserting things about, then it is also redundant, I didn't see why there needs to be a comment saying "this is the last command before we begin to assert things", you can just see that it is the last command before asserting things and/or that the assertions are all about the response to that command.

katago_server/distributed_efforts/tests/test_api.py
lightvector: Still need to experiment with .save() and deal with tests with uploading networks and games though! :)

katago_server/games/managers/rating_game_pandas_queryset.py
lightvector: I'd like to understand why, at some point though...

PR #19, opened 2020-08-17
More fixes

lightvector:
Updates the network selection for training to just simple - just use the latest network. Probably actually a good idea for improving game variety, minimizes the amount of latency and wasted delay from rating/gating.

Resolves various todos and fixes bayes elo race condition, adds some more validation and tests, adds configurable way to enable/disable networks from training or rating games.


lightvector:
Mostly making this pull request for more visibility, will go ahead and merge in a few days if there aren't strong objections.

PR #21, opened 2020-08-24
Make npm installation of gifsicle work

lightvector:
This seems to make it work. We need the essential build stuff and autoconf before we try to get npm working.


lightvector:
Going ahead to merge this, since it seems to work in testing and I don't see how it would be objectionable.

PR #30, opened 2020-10-04
Larger limits for requests

lightvector:
@tychota - do you think this is reasonable?

I ran into an error for this in the admin panel when I tried to select and delete about 10000 startposes all at once.


lightvector:
Edit: actually I mean delete 1000 at once (a little more than 1000). The default limit in django appears to be just 1000.

PR #31, opened 2020-10-04
Delete training data and sgfs when games are deleted from the database

lightvector:
I think we very definitely need this behavior for cleaning up deleted games and sgfs. Imagine: we find someone uploading some low-quality data due to a bug or accident or even adversarial activity: we delete the games in admin, and we think problem is solved.

Nope! The training will still be using that low-quality data, unless we have something like this. So this change is important.

@tychota - I'm submitting this for review but I think it's not ready yet because of the plan to support Google Cloud Storage. I partly understand now the idea for how to be flexible for the storage class used, but what is a good way to write deletion code like this to also be flexible?


tychota:
> I think we very definitely need this behavior for cleaning up deleted games and sgfs. Imagine: we find someone uploading some low-quality data due to a bug or accident or even adversarial activity: we delete the games in admin, and we think problem is solved.
> 
> Nope! The training will still be using that low-quality data, unless we have something like this. So this change is important.

Makes full sense

> @tychota - I'm submitting this for review but I think it's not ready yet because of the plan to support Google Cloud Storage. I partly understand now the idea for how to be flexible for the storage class used, but what is a good way to write deletion code like this to also be flexible?

You can just use the delete method on fields:

```python

@receiver(post_delete, sender=TrainingGame)
def auto_delete_file_on_delete(sender, instance, **kwargs):
    if instance.sgf_file:
        instance.sgf_file.delete()
    # ...

```

It will automatically delete based on the storage

https://github.com/jschneier/django-storages/blob/master/storages/backends/gcloud.py#L163-L164

lightvector:
Thank you for the help about this and about the file storage generality. Does this change look good to you now?

lightvector:
I did a little bit of refactoring of the file storage method in that second commit - file storage is now dehardcoded, and although it is still using the local disk, it should now be only a short step to change it to use cloud storage, only needing to swap out the production.py and to specify the relevant buckets and such in .envs.

tychota (approved):

PR #37, opened 2020-10-21
More basic frontend

lightvector:
Work in progress. Adding contributions page and various improvements. This is incomplete - need to work out performance issues in order to be able to display more useful game data totals and such. Will try manifested views in the next few days. 

I'm opening this pull request anyways to provide some visibility on the refactors and changes so far, even though this is still in-progress.


lightvector:
I pushed a bunch of optimizations to performance, including a using custom template tags to build the URLs for different parts of the sites without relying on `url` which uses Django's `reverse` which seems very slow. We still use `url` within many casual parts of the site, but within the loops within the tables, which get called hundreds of times, we try to avoid it.

I'm curious what you think.

lightvector:
The prefetching is also a big improvement. On my local test server, collectively these optimizations improve performance experienced by the browser quite noticeably - it feels much less laggy to navigate around.

src/urls.py
```diff
@@ -100,28 +105,28 @@
...
+            "frontend/training-games/<run>/by-network/<network>/<page>/",
+            GamesListByNetworkView.as_view(),
             {"kind": "training"},
-            name="training_games_list"
+            name="training_games_list_by_network"
         ),
         path(
-            "frontend/rating-games/<run>/<network>/",
-            GamesListView.as_view(),
+            "frontend/rating-games/<run>/by-network/<network>/",
```
tychota, line 133: Iwill say that the url that makes more sense is 

`frontend/network/<network>/rating-games/` (not sure if frontend is nice to have in the url, by the way.)

src/frontend/views/games_view.py
```diff
@@ -45,8 +45,8 @@ def get_context_data(self, **kwargs):
       return context
 
 
-class GamesListView(ListView):
-  template_name = "pages/games_list.html"
+class GamesListByNetworkView(ListView):
```
tychota, line 37: Maybe we should split the views into a directory of views ?

src/frontend/views/games_view.py
```diff
@@ -45,8 +45,8 @@ def get_context_data(self, **kwargs):
...
+  template_name = "pages/games_list_by_network.html"
   paginate_by = 100
```
tychota, line 39: By the way do you know the diff between offset and cursor pagination ?

https://use-the-index-luke.com/no-offset

I think we should use cursor based pagination everywhere with a reasonable size if we want to improve performance 

https://github.com/photocrowd/django-cursor-pagination

src/frontend/templates/pages/game_network_groups.html
```diff
@@ -33,8 +33,8 @@ <h1 class="title">{% blocktrans with run_name=run.name %}Games for {{run_name}}
   {% for network in network_list %}
   <tr>
     <td> {{ network.name }} </td>
-    <td> <a href="{% url 'training_games_list_by_network' run.name network.name 1 %}"> {% trans "Training Games" %} </td>
-    <td> <a href="{% url 'rating_games_list_by_network' run.name network.name 1 %}"> {% trans "Rating Games" %} </td>
+    <td> <a href="{% url 'training_games_list_by_network' run.name network.name %}"> {% trans "Training Games" %} </td>
```
tychota: what was the 1 ?

src/frontend/templates/pages/games_list_by_network.html
```diff
@@ -18,31 +18,7 @@ <h1 class="title">{% blocktrans with network_name=network.name %} Training Games
...
-    <a class="mx-1" href="{% url list_url_name run.name network.name page_obj.next_page_number %}"> {% trans "Next" %} <i class="fas fa-angle-right"></i></a>
-    <a class="mx-1" href="{% url list_url_name run.name network.name page_obj.paginator.num_pages %}"> {% trans "Last" %} <i class="fas fa-angle-double-right"></i></a>
-    {% else %}
-    <span class="mx-1"> {% trans "Next" %} <i class="fas fa-angle-right"></i></span>
-    <span class="mx-1"> {% trans "Last" %} <i class="fas fa-angle-double-right"></i></span>
-    {% endif %}
-  </span>
-</div>
-{% endif %}
+{% include 'pagination.html' with other_get_params='' page_obj=page_obj %}
```
tychota, line 23: Nice

src/frontend/templates/pages/games_list_by_user.html
```diff
@@ -0,0 +1,68 @@
...
+<table class="table">
+  <tr>
+    <th> {% trans "Game ID" %} </th>
+    <th> {% trans "Uploaded At" %} </th>
+    {% if kind == "rating" %}
+    <th> {% trans "Black" %} </th>
+    <th> {% trans "White" %} </th>
+    {% else %}
+    <th> {% trans "Network" %} </th>
+    {% endif %}
```
tychota, line 34: That feeels weird for me.

ideally I think we could have two tables, one for "rating", one for "training games" and just "if kind" one in the vue to include the good table rather than "if kind" x times in the table.

That would increase cohesion and reduce complexity.

src/frontend/templates/pagination.html
```diff
@@ -1,26 +1,29 @@
...
-{% if is_paginated %}
-<nav class="pagination is-centered">
-  {% if page_obj.has_previous %}
-    <a class="pagination-previous" href="?page=
-        {{ page_obj.previous_page_number|stringformat:"d" }}{{ getvars }}{{ hashtag }}">{% trans "Previous" %}</a>
-  {% else %}
-    <a class="pagination-previous" disabled>{% trans "Previous" %}</a>
-  {% endif %}
+{% if page_obj.has_previous or page_obj.has_next %}
+<div class="pagination">
```
tychota, line 4: nice

src/frontend/views/contributions_view.py
```diff
@@ -0,0 +1,24 @@
...
+from src.apps.users.models import User
+
+from . import view_utils
+
+class ContributionsView(ListView):
+  template_name = "pages/contributions.html"
+  context_object_name = "user_list"
+
+  def get_queryset(self):
+    view_utils.set_current_run_or_run_from_url_for_view_get_queryset(self)
```
tychota, line 15: why isn't the run explicit ?

Idealy we would have diiferents views:
- one for `fontend/run/<run>/contributions` where run is the run to display and may be the current run (by default, links redirect to `run/test/networks` when `test` is the active run
- one for `fontend/contributions` which don't filter by active run and display the contribution for all runs (no filter)

src/frontend/views/games_view.py
```diff
@@ -74,6 +59,34 @@ def get_context_data(self, **kwargs):
...
+
+  def get_queryset(self):
+    self.run = get_object_or_404(Run, name=self.kwargs["run"])
+    self.user = get_object_or_404(User, username=self.kwargs["user"])
+
+    if self.kwargs["kind"] == "training":
+      games = TrainingGame.objects.filter(submitted_by=self.user,run=self.run).order_by("-created_at")
+    else:
+      games = RatingGame.objects.filter(submitted_by=self.user,run=self.run).order_by("-created_at")
+    return games
```
tychota: Also not sure weither it is bettr to have this kind logic vs having two views.

```py
class TrainingGamesListByUserView(ListView): 
  template_name = "pages/training_games_list_by_user.html"
  paginate_by = 100

  def get_queryset(self):
    self.run = get_object_or_404(Run, name=self.kwargs["run"])
    self.user = get_object_or_404(User, username=self.kwargs["user"])
    return TrainingGame.objects.filter(submitted_by=self.user,run=self.run).order_by("-created_at")
``` 

I think by putting if we decrease cohesion and we are less clear.

src/frontend/views/games_view.py
```diff
@@ -74,6 +59,34 @@ def get_context_data(self, **kwargs):
...
+      context["run"] = self.run
+      context["user"] = self.user
+      context["kind"] = self.kwargs["kind"]
+      if self.kwargs["kind"] == "training":
+        context["list_url_name"] = "training_games_list_by_user" # urls.py
+        context["sgfplayer_url_name"] = "sgfplayer_training" # urls.py
+      else:
+        context["list_url_name"] = "rating_games_list_by_user" # urls.py
+        context["sgfplayer_url_name"] = "sgfplayer_rating" # urls.py
+      return context
```
tychota: Same than above, we are having a lot of if, in view, template. We should have two views, two templates and no if, then extract the similar parts.

src/frontend/views/view_utils.py
```diff
@@ -0,0 +1,41 @@
...
+def set_current_run_or_run_from_url_for_view_get_queryset(view):
+  """Helper for implementing get_queryset in a view that cares about the run.
+
+  Sets view.run to:
+    * The run specified by the url (from view.kwargs)
+    * Or else, the current run if there is a current active run
+    * Or else, None
+
+  Sets view.viewing_current_run to False if the run was in the url, else True.
+  """
```
tychota, line 14: Also I start being less clear about this.

- having no run in url should means: "I want to see data from all runs" (eg my overall contributions).
- else we should always have a run in url, explicitly.

src/frontend/templates/base.html
```diff
@@ -51,23 +51,33 @@
...
+          <a class="navbar-item is-active" href="https://github.com/lightvector/KataGo/releases">
+            {% trans "Releases" %}
+          </a>
+          <a class="navbar-item is-active" href="https://github.com/lightvector/KataGo">
+            {% trans "GitHub (KataGo)" %}
+          </a>
+          <a class="navbar-item is-active" href="https://github.com/katago/katago-server">
+            {% trans "GitHub (Webserver)" %}
+          </a>
+
```
tychota, line 64: nice.

is-active vs is-hidden-desktop-only ?

requirements/local.txt
```diff
@@ -27,6 +27,7 @@ pre-commit==2.4.0  # https://github.com/pre-commit/pre-commit
 factory-boy==2.12.0  # https://github.com/FactoryBoy/factory_boy
 
 django-debug-toolbar==2.2  # https://github.com/jazzband/django-debug-toolbar
+django-debug-toolbar-template-profiler==2.0.1 # https://pypi.org/project/django-debug-toolbar-template-profiler/
```
tychota, line 30: Nice

src/settings/local.py
```diff
@@ -38,20 +38,35 @@
...
 # https://django-debug-toolbar.readthedocs.io/en/latest/installation.html#internal-ips
 INTERNAL_IPS = ["127.0.0.1", "10.0.2.2"]
-if env("USE_DOCKER") == "yes":
-    import socket
-
-    hostname, _, ips = socket.gethostbyname_ex(socket.gethostname())
-    INTERNAL_IPS += [ip[:-1] + "1" for ip in ips]
+if env("USE_DOCKER") == "yes" and DEBUG:
+    # https://stackoverflow.com/a/49818040
+    INTERNAL_IPS = type(str('c'), (), {'__contains__': lambda *a: True})()
```
tychota, line 69: Nice

tychota: Why wasn't https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/config/settings/local.py#L64-L72 working ?

src/apps/games/tests/test_game.py
```diff
@@ -134,7 +134,7 @@ def setup_method(self):
         self.bad_games.extend(self.create_games_with_defaults(board_size_x=45))
         self.bad_games.extend(self.create_games_with_defaults(board_size_y=2))
         self.bad_games.extend(self.create_games_with_defaults(board_size_y=45))
-        self.bad_games.extend(self.create_games_with_defaults(handicap=25))
+        self.bad_games.extend(self.create_games_with_defaults(handicap=2500))
```
tychota, line 137: Ah yes !

tychota: Note to myself, setup a CI to run the tests.

tychota: Maybe we should test the view: eg https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/%7B%7Bcookiecutter.project_slug%7D%7D/users/tests/test_views.py

.envs/.local/.postgres
```diff
@@ -2,6 +2,6 @@
 # ------------------------------------------------------------------------------
 POSTGRES_HOST=postgres
 POSTGRES_PORT=5432
-POSTGRES_DB=katago_server
+POSTGRES_DB=katago_server_db
```
tychota, line 5: ok

src/apps/trainings/models/network.py
```diff
@@ -128,6 +129,18 @@ def elo_uncertainty(self):
...
+    @property
+    def download_link_html(self):
+        if self.model_file:
+            return f'<a href="{self.model_file.url}">[{download_translated}]</a>'
+        return ""
+    @property
+    def zip_download_link_html(self):
+        if self.model_zip_file:
+            return f'<a href="{self.model_zip_file.url}">[{download_translated}]</a>'
+        return ""
```
tychota, line 141: Looks weird for me to put template in model. It breaks the single responsability principle

For me that should rather be a template tag `obj.model_file|download_link`

src/frontend/templatehelpers/templatetags/custom_time_formatting.py
```diff
@@ -0,0 +1,18 @@
+from django import template
+
+register = template.Library()
+
+# Peformance hack for converting times to a fixed ISO-like format noticeably faster
```
tychota, line 5: Nice to know. 

Do you know the performance impact of just this ?

src/frontend/templatehelpers/templatetags/custom_url_tags.py
```diff
@@ -0,0 +1,28 @@
+from django import template
+
+register = template.Library()
+
+# These are performance hacks that directly build a relative URL for certain pages
+# instead of relying on django reverse(), which seems to be very slow when called
+# hundreds of times, once for each different item in a lst.
+
+@register.simple_tag
+def training_games_list_by_network(run_name, network_name):
```
tychota, line 10: Nice to know.

Do you know the performance impact of just this ?

src/frontend/views/games_view.py
```diff
@@ -39,10 +43,10 @@ def get_queryset(self):
     self.network = get_object_or_404(Network, name=self.kwargs["network"])
 
     if self.kwargs["kind"] == "training":
-      games = TrainingGame.objects.filter(white_network=self.network).order_by("-created_at")
+      games = TrainingGame.objects.filter(white_network=self.network).order_by("-created_at").prefetch_related("submitted_by")
```
tychota, line 46: Nice to know.

Do you know the performance impact of just this ?

I think this is the real pain / culprit and is alone responsible of 99% of the slowness (much more that time format or reverse.

tychota (approved):

lightvector:
Added materialized view for counting games by network. I'll do another one for counting games by user soon.

Also, right now the views are never refreshed. Will add a celery task that refreshes them as well.

Question: Is there a way to make a celery task created and initialized and active with a default period (e.g. 2 minutes) upon server build and migrate and normal startup? When I set up the production server I had to manually add tasks and intervals for things like the Bayes Elo.

lightvector:
I will address your review comments and fix the things you pointed out in the next days as well, and investigate cursor pagination.

src/frontend/templates/pages/game_network_groups.html
lightvector: The 1 was the page number. It is unnecessary now since we are using the GET arguments for pagination which is better and gives us the page 1 for free by default.

src/apps/games/tests/test_game.py
lightvector: A CI to run the tests would be great!

src/apps/trainings/models/network.py
lightvector: You are right, thanks. I fixed it and used a template filter instead.

src/frontend/views/games_view.py
lightvector: It cut down the load time of some of the pages by about 70%-90%. Yay! :)

src/settings/local.py
lightvector: I don't know why it wasn't working, and I don't understand how the ips work well enough to know what was supposed to be correct or not, or what my ips were inside or outside of socker. But it definitely was not working for me, whereas the new replacement code is working for me.

src/frontend/templates/base.html
lightvector: I don't know what is-active actually does, I only did it to mimic the existing menu items. :)

src/frontend/templatehelpers/templatetags/custom_time_formatting.py
lightvector: Fairly small, like 10ms or something for 500 networks. But it's a pretty simple change, I think, and doesn't break any abstraction boundaries.

src/frontend/templatehelpers/templatetags/custom_url_tags.py
lightvector: This one is like 50 or 70 milliseconds or something like that, when there is a lot of items in the table. Apparently reverse URL lookup is very slow.

src/frontend/views/games_view.py
lightvector: What do you mean? What is a directory of views?

src/urls.py
lightvector: Thanks, got rid of "frontend".
What would you propose for the replacement for "training-games/<run>/by-user/<user>/"  then? We would want to keep it consistent.

src/frontend/templates/pages/games_list_by_user.html
lightvector: Thanks, I will consider it. Right now it is the *only* thing in the table that varies between rating and training - is it really good style to duplicate all the code for 11 other identical columns to save on 1 if statement?

src/apps/games/models/game_count_views.py
```diff
@@ -16,49 +16,56 @@ class GameCountByNetwork(pg.MaterializedView):
...
+        SELECT
+        black_network_id as network_id,
+        count(*) as total_num_training_games,
+        sum(num_training_rows) as total_num_training_rows
+        FROM games_traininggame
+        GROUP BY black_network_id
+      ) trainingcounts
+      FULL OUTER JOIN
       (
+        SELECT network_id, sum(num_games) as total_num_rating_games, sum(score) as total_rating_score
```
tychota, line 41: CTE (having a select inside a query) is consuming some memory.

We may change the sql later (optimize it).

src/apps/games/tests/test_game.py
```diff
@@ -177,10 +174,228 @@ def teardown_method(self):
...
+    def test_game_validations(self):
         for game in self.good_games:
             game.full_clean()
         for game in self.bad_games:
             with pytest.raises(ValidationError):
                 game.full_clean()
         pass
+
+
+class TestMaterializedGameViews:
```
tychota, line 186: nice

src/apps/games/tasks/refresh_materialized_game_views.py
```diff
@@ -0,0 +1,13 @@
...
+from src.apps.games.models import GameCountByNetwork, GameCountByUser
+
+
+@celery_app.task()
+def refresh_materialized_game_views():
+    """
+    Refresh all the materialized views for games app
+    :return:
+    """
+    GameCountByNetwork.refresh(concurrently=True)
```
tychota, line 12: nice

tychota (approved):

src/apps/games/tests/test_game.py
lightvector: :)

src/apps/games/tasks/refresh_materialized_game_views.py
lightvector: Is there a way to automatically set up the celery task with a default period upon migrate or do I just do that manually in admin panel?

src/frontend/views/view_utils.py
lightvector: I left it this way for networks since for networks it make a lot of sense to default to current run. But for contributors I took your suggestion and greatly simplified to to just view all games from all runs. Thanks!

src/frontend/views/contributions_view.py
lightvector: I thought about and agree, I went with just making the contributors page look at all games from all runs.

src/urls.py
lightvector: Okay I set a new URL scheme that is extremely similar to what you proposed. Thanks!

src/frontend/views/games_view.py
lightvector: I think I will leave this as todo for future work, since the current thing works okay. This is frontend-only, so that makes it easy to test and change in the future.

PR #39, opened 2020-10-23
Dechardcode a few emails and domains

lightvector:
A big thing left to dehardcode is a bunch of stuff in traefik.toml that is hardcoded to katago.tycoach.me, as well as your tychota email address.

I don't know how to dehardcode it yet. @tychota - any suggestions?


lightvector:
Okay I attempted to dehardcode traefik and update a bunch of other things. I really don't like how the .envs directory is not visible by default when you ls even though it is part of the repo and is important. So I moved them to not have a dot in the name and I also added examples for the envs for production and updated the readme.

@tychota - Can you read over especially the traefik docker-compose part and tell me if it looks reasonable? I have *not* tried it yet and I am not sure it is correct, but if it looks good to you I can try it out on the "prod" server and see if anything breaks. I don't think I can test it purely on the local server because traefik is only part of production.yml.

lightvector:
Okay, it is fixed and working now. The problem was that we were missing the loadbalancer option (because we deleted them) to specify the port. This is fixed now.

And I updated the readmes and other documentation for environment variables. And I also restored the toml file because apparently tls options are NOT possible to specify through docker labels. But I renamed it to "traefik-dynamic.toml" because if it is named "traefik.toml" then traefik attempts to read a static config from it and fails.

Everything seems working now. If it a final skim looks good to you too, then let me know and we can merge.

production.yml
```diff
@@ -32,7 +34,22 @@ services:
...
+      # Incoming http to our domain name via websecure entrypoint gets routed to django, modulo priority
+      traefik.http.routers.djangoroute.rule: "Host(`${TRAEFIK_ROUTER_DOMAIN_NAME:?traefik_router_domain_name_missing}`)"
+      traefik.http.routers.djangoroute.entrypoints: "websecure"
+      traefik.http.routers.djangoroute.tls.certresolver: "letsencrypt"
+      traefik.http.routers.djangoroute.priority: 10
+      # Django internally listens on port 5000 by default - presumably we could configure django to listen on port 80 or 443,
+      # but we can also just tell traefik to listen here.
+      traefik.http.services.djangoroute.loadbalancer.server.port: 5000
+      # Define the "csrf" middleware and make djangoroute use it
+      traefik.http.middlewares.csrf.headers.hostsProxyHeaders: "X-CSRFToken"
```
tychota, line 48: Maybe this line should ve in traefik section

tychota (approved):
Well done

production.yml
lightvector: It works this way according to the API (shows middleware successfully) but it does not work if I delete it from both django and nginx and put it in traefik labels. So apparently it needs to be duplicated locally for each container that uses it.

PR #40, opened 2020-10-25
Finish frontend

lightvector:
This is a work-in-progress branch to finish a basic implementation of the frontend.

To start with, cleaning up a few things and adding support for stats about the networks which will be useful for graph on the front page, as well as of general interest to anyone making their own plots.


tychota (approved):

lightvector:
Fancy front page graph using d3. Woooo. :)

I deleted the about page since I didn't really know what should be on there, we already have links to the githubs and the front page can describe the basics once we go public.

We can always add an faq page or whatever else we need later, if we need it.

lightvector:
I think this completes a usable frontend, except for some final description and instructions for how to contribute, which we will fill in later.

PR #42, opened 2020-10-27
Frontend work

lightvector:
@tychota - This addresses a lot of the suggestions you mentioned in discord, as well as overhauls the method for viewing the old runs.

In rough order that you listed things. 

* Elo graph at top - Yes, Elo graph is moved a bit closer to the top now.
* About page - no, I think we should still keep about deleted and just have on the front page directions to contribute, so it is more prominent.
* Reading the data numbers - I tried bolding and I didn't like it, so I didn't do this for now. The problem is that the sentence is so compact anyways that like 1/2 or 2/3 of the sentence is bolded, depending on how you do it - and if too much is bold, it is as if nothing is bold. And bolding only parts of it is also inconsistent with some of the later numbers below too.
* Thousands separators for data numbers - is there a way to do this without being a total mess, given that it is also inside lots of blocktrans, and blocktrans doesn't allow for anything except raw variables? (no filters, etc). I didn't do this yet and won't consider it a high priority given how messy it might be.
* Strongest net download - I added a link, thanks.
* Spacing - I adjusted it a little, thanks.
* Last week, last month, etc - I stick with last week for now. I'm not sure how I feel about calling it a "leaderboard" explicitly. DIdn't do this.
* Too long - I renamed "Training rows generated" to just "Training Rows" which slightly shortens it. I don't like having a compact notation though like you suggested - but if we need it to be shorter we could just make it "games" in the table entries instead of "training games" and "rating games". But leave the table header the same.
* No, let's not document here an explicit percentage of game that are rating games. That's sort of just implied, and people who contribute will notice it. If we later have an FAQ, it could go there.
* Network names - I left the run name on because I want to err heavily on the side of discouraging people from mixing things up with other runs and getting confused. I tried leaving it off briefly, and I didn't like the result. Can change it later if needed.
* s and d - I added a tooltip for that on the table header.
* Date formatting - I left it UTC since this is simpler, and also means that if people try to communicate to each other the network by date (e.g. the network uploaded on date xxxx) there is no chance of confusion if they all see the same date and time.
* Download as a header - Thanks, I fixed it.
* Link the games in the networks table - Maybe that makes the networks table quite long? I sorta like the current format. What makes you confused about it right now? If you want to see games, you click "games" in the menu. If you want to download networks, you click "networks" in the menu. I don't see a need to link long tables of games for each network and make the networks table too big - I expect that this page will frequently be used by users who just want to download a particular network version.
* Other runs - I completely overhauled the method. There are no tables at the bottom any more, instead there is one simple page indexing the runs, and a standardized template that every page includes that says whether it for current run or older run, and links to the index page or to the current run page as necessary. This also simplifies a lot of the logic that I think you didn't like.


tychota:
> Reading the data numbers - I tried bolding and I didn't like it, so I didn't do this for now. The problem is that the sentence is so compact anyways that like 1/2 or 2/3 of the sentence is bolded, depending on how you do it - and if too much is bold, it is as if nothing is bold. And bolding only parts of it is also inconsistent with some of the later numbers below too.

Have you tried `text <em> important </em>`

lightvector:
> > Reading the data numbers - I tried bolding and I didn't like it, so I didn't do this for now. The problem is that the sentence is so compact anyways that like 1/2 or 2/3 of the sentence is bolded, depending on how you do it - and if too much is bold, it is as if nothing is bold. And bolding only parts of it is also inconsistent with some of the later numbers below too.
> 
> Have you tried `text <em> important </em>`

You mean to make them italics by default? Does that help readability? I'm not sure I like that either. :)

tychota:
> Thousands separators for data numbers - is there a way to do this without being a total mess, given that it is also inside lots of blocktrans, and blocktrans doesn't allow for anything except raw variables? (no filters, etc). I didn't do this yet and won't consider it a high priority given how messy it might be.

I think django handle it either with https://docs.djangoproject.com/fr/3.1/ref/settings/#std:setting-THOUSAND_SEPARATOR or https://docs.djangoproject.com/en/3.1/ref/contrib/humanize/

lightvector:
The same problem remains - if you emphasize the phrases "X rating games" and whatever in that sentence, then there is actually *more* that is emphasized in the sentence than isn't emphasized. It looks pretty weird to me, and with so much emphasis, it didn't feel like it helped visually highlight anything important.

I suspect thousands separators would help the most here, if it could be done cleanly across the site.

lightvector:
> > Thousands separators for data numbers - is there a way to do this without being a total mess, given that it is also inside lots of blocktrans, and blocktrans doesn't allow for anything except raw variables? (no filters, etc). I didn't do this yet and won't consider it a high priority given how messy it might be.
> 
> I think django handle it either with https://docs.djangoproject.com/fr/3.1/ref/settings/#std:setting-THOUSAND_SEPARATOR or https://docs.djangoproject.com/en/3.1/ref/contrib/humanize/

Should I be afraid of horror stories online where setting this sitewide causes bugs and problems when formatting ID numbers when building links, and things like that?

tychota:
> About page - no, I think we should still keep about deleted and just have on the front page directions to contribute, so it is more prominent.

> Last week, last month, etc - I stick with last week for now. I'm not sure how I feel about calling it a "leaderboard" explicitly. DIdn't do this.

> No, let's not document here an explicit percentage of game that are rating games. That's sort of just implied, and people who contribute will notice it. If we later have an FAQ, it could go there.

> Network names - I left the run name on because I want to err heavily on the side of discouraging people from mixing things up with other runs and getting confused. I tried leaving it off briefly, and I didn't like the result. Can change it later if needed.

Ok for me

tychota:
> Should I be afraid of horror stories online where setting this sitewide causes bugs and problems when formatting ID numbers when building links, and things like that?

Not sure. The second link is a tag so i will say it is explicit.

tychota:
> Link the games in the networks table - Maybe that makes the networks table quite long? I sorta like the current format. What makes you confused about it right now? If you want to see games, you click "games" in the menu. If you want to download networks, you click "networks" in the menu. I don't see a need to link long tables of games for each network and make the networks table too big - I expect that this page will frequently be used by users who just want to download a particular network version.

I think it is a bit confusing to see network list in two places, with different info.
I'm a bit lost so maybe user will be too (or maybe not :) )

src/frontend/static/css/elo-graph.css
```diff
@@ -1,20 +1,14 @@
 #eloGraph {
   width: 90%;
+  /* Disable double click to zoom on mobile */
+  touch-action: manipulation;
```
tychota, line 4: nice

src/frontend/static/css/tooltips.css
```diff
@@ -0,0 +1,24 @@
+
+/* Makes "title" tootlip attribute work on mobile on a click as well.
+   Requires element to also have a tabindex. */
+@media (pointer: coarse), (hover: none) {
```
tychota, line 4: nice

src/frontend/templates/pages/game_network_groups.html
```diff
@@ -46,16 +46,16 @@ <h1 class="title">{% blocktrans with run_name=run.name %}Games for {{run_name}}
...
-<h1 class="title"> {% trans "Games for Other Runs" %} </h1>
+<h1 class="title"> {% trans "Games for Older Runs" %} </h1>
 <ul>
 <table class="table">
   <tr>
     <th> {% trans "Run Name" %} </th>
   </tr>
 
-  {% for run in other_runs_to_show %}
+  {% for other_run in other_runs_to_show %}
```
tychota: nice

src/frontend/templates/pages/home.html
```diff
@@ -1,448 +1,14 @@
...
 {% load custom_translation %}
 
 {% block extra_html_head_stuff %}
-<link type="text/css" href="/static/css/elo-graph.css" rel="stylesheet" />
 <link type="text/css" href="/static/css/tooltips.css" rel="stylesheet" />
 
-<script type="text/javascript" src="/static/js/d3.v6.min.js"></script>
-
 {% if run %}
-{% url 'networks-for-elo-list' as networks_for_elo_list_url %}
```
tychota, line 12: nice

src/frontend/templates/run_stats_head.html
```diff
@@ -0,0 +1,440 @@
...
+        .domain(computeYDomainFromXDomain(xScale.domain()))
+        .range([graphSpaceHeight, 0]);
+    var yAxis = d3.axisLeft(yScale).tickSize(-graphSpaceWidth*1.0).ticks(20);
+    var yAxisContent = graphSpace.append("g");
+    yAxis(yAxisContent);
+    var unzoomedYDomainMin = yScale.domain()[0];
+    var unzoomedYDomainMax = yScale.domain()[1];
+
+    // Colors
+    var colorScale = d3.scaleSequential(d3.interpolateWarm).domain([0,networkSizes.length-1]);
```
tychota, line 174: Not important: on actual site i find the color not going well with the reste of the site

Maybe we can take bulma colors: https://bulma.io/documentation/overview/colors/

src/frontend/templates/run_stats_head.html
```diff
@@ -0,0 +1,440 @@
...
+        console.log(networkSize);
+        d3.selectAll(".dataUncertainty_" + networkSize).attr("stroke-width",curStrokeWidth + largeStrokeWidthBonus);
+        d3.selectAll(".dataPoint_" + networkSize).attr("r",curDataRadius + largeRadiusBonus);
+      })
+      .on("mouseleave", function(mouseEvent,networkSize) {
+        d3.selectAll(".dataUncertainty_" + networkSize).attr("stroke-width",curStrokeWidth);
+        d3.selectAll(".dataPoint_" + networkSize).attr("r",curDataRadius);
+      });
+
+    console.log(networkSizes);
```
tychota, line 433: Maybe we should remove console.log

src/frontend/templates/run_stats_head.html
```diff
@@ -0,0 +1,440 @@
...
+        return 0;
+      if(aIsRandom)
+        return -1;
+      if(bIsRandom)
+        return 1;
+      return collator.compare(a,b);
+    };
+    networkSizes = Array.from(networkSizes).sort(compare);
+
+    console.log(networkSizes);
```
tychota, line 32: Maybe remove networks

src/frontend/templates/run_stats_head.html
```diff
@@ -0,0 +1,440 @@
...
+      aIsRandom = a.toLowerCase() == "random";
+      bIsRandom = b.toLowerCase() == "random";
+      if(aIsRandom && bIsRandom)
+        return 0;
+      if(aIsRandom)
+        return -1;
+      if(bIsRandom)
+        return 1;
+      return collator.compare(a,b);
+    };
```
tychota, line 29: Another solution is having django outputing directly data. 
I think you solution is cleaner.

src/frontend/views/home_view.py
```diff
@@ -12,74 +9,14 @@ class HomeView(TemplateView):
 
   def get_context_data(self, **kwargs):
     context = super().get_context_data(**kwargs)
-    context["top_recent_user_list"] = (
```
tychota, line 15: woah, nice clean

src/frontend/templates/pages/networks.html
```diff
@@ -32,15 +32,19 @@ <h1 class="title">{% blocktrans with run_name=run.name %}Networks for {{run_name
 {% elif run and network_list|length > 0 %}
 
 {% blocktrans with katago_releases_url="https://github.com/lightvector/KataGo/releases" %}
-Stats and download links for all the neural networks from the current run. Downloaded nets can be used directly with <a href="{{katago_releases_url}}">KataGo</a>.
+Stats and download links for all the neural networks from the current run.
```
tychota: nice description

src/frontend/templates/pages/contributions.html
```diff
@@ -30,13 +40,14 @@ <h5 class="is-size-5"> {% trans "Top contributors for KataGo distributed trainin
...
+{% endif %}
 
-<h5 class="is-size-5"> {% trans "All contributors for KataGo distributed training across all time:" %} </h5>
+<h5 class="is-size-5 mt-4"> {% trans "All contributors for KataGo distributed training across all time:" %} </h5>
 
 <table class="table mt-2">
   <tr>
     <th> {% trans "User" %} </th>
-    <th class="has-text-right"> {% trans "Training Rows Generated" %} </th>
+    <th class="has-text-right"> {% trans "Training Rows" %} </th>
```
tychota, line 50: nice

tychota (approved):

src/frontend/views/home_view.py
lightvector: :)

src/frontend/templates/run_stats_head.html
lightvector: Thanks.

src/frontend/templates/pages/home.html
lightvector: Factored this out so that we could reuse it both on home page and on runs view.

src/frontend/templates/run_stats_head.html
lightvector: Good catch, fixed, thanks.

src/frontend/templates/run_stats_head.html
lightvector: Thanks, fixed.

PR #44, opened 2020-11-03
Write helm kubernetes yml files

lightvector:
Attempt to write yml files for kubernetes. Wooo!

@tychota - When you get the chance, please let me know what you think. I have a few questions about various things - the expected workflow for using these files, versioning and tagging and upload of images, how secrets work, how internal services communication and hostname works for redis and celery, but we can address these in an audio chat when you have time.

Redis is missing from this push, but I think mostly everything else is here aside from secrets?


lightvector:
Also I'm not sure that this is the right place to put these - maybe they belong in another repo. But given that "docker compose" configuration lives in this same repo, and kubernetes yml is supposed to be a replacement for "docker compose", maybe it makes sense to put it here too? So that if we need to update it (for example django needs a new environment variable or changes in its start command) we can update both docker-compose and kubernetes in a synchronized way?

Not sure. This is a part of the workflow I am fuzzy on, let me know. :)

tychota:
> Also I'm not sure that this is the right place to put these - maybe they belong in another repo. But given that "docker compose" configuration lives in this same repo, and kubernetes yml is supposed to be a replacement for "docker compose", maybe it makes sense to put it here too? So that if we need to update it (for example django needs a new environment variable or changes in its start command) we can update both docker-compose and kubernetes in a synchronized way?

I would say that they are more decoupled that docker-compose thought.

Eg sometimes docker-compose say `build: ./relative/path/to/dockerfile` but here we separate this into two opérations. Building image is not running them.

kubernetes/celery-beat/deployment.yml
```diff
@@ -0,0 +1,51 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: celery-beat-deployment
```
tychota, line 4: [Importance 2/10]

Maybe we should create a namespace `namespace: website` and have the namespace created in another file.

Having some namespace makes it easy to say for instance thatpod inside  `website` should have a service mesh attached on.

kubernetes/celery-beat/deployment.yml
```diff
@@ -0,0 +1,51 @@
...
+  labels:
+    app: celery-beat
+spec:
+  replicas: {{.Values.celeryBeat.replicaCount}}
+  strategy:
+    type: Recreate
```
tychota, line 10: [Importance 5/10]

Recreate (in opposition to RollingUpdate) destroy all then restart all.
This way you effectivly cut the service for a few seconds. Is it expected ?

kubernetes/celery-beat/deployment.yml
```diff
@@ -0,0 +1,51 @@
...
+      app: celery-beat
+  template:
+    metadata:
+      labels:
+        app: celery-beat
+    spec:
+      containers:
+      - name: celery-beat
+        image: {{.Values.django.image.repo}}:{{.Values.django.image.tag}}
+        command: [ "/entrypoint", "/start-celerybeat"]
```
tychota, line 22: [Importance 2/10]

Entrypoint doesn't work in kubernetes (as you noticed).

Maybe we can rework images to have plain script instead of two scripts
Eg just

/start-celerybeat

kubernetes/celery-beat/deployment.yml
```diff
@@ -0,0 +1,51 @@
...
+        envFrom:
+        - secretRef:
+            name: django-env-secrets
+        - secretRef:
+            # POSTGRES_USER and POSTGRES_PASSWORD
+            name: postgres-env-secrets
+        env:
```
tychota, line 29: You can also use env from but from configmap, this time

https://kubernetes.io/docs/concepts/configuration/configmap/

kubernetes/celery-beat/deployment.yml
```diff
@@ -0,0 +1,51 @@
...
+  template:
+    metadata:
+      labels:
+        app: celery-beat
+    spec:
+      containers:
+      - name: celery-beat
+        image: {{.Values.django.image.repo}}:{{.Values.django.image.tag}}
+        command: [ "/entrypoint", "/start-celerybeat"]
+        envFrom:
```
tychota, line 23: Nice, that is exactly what i had in mind

kubernetes/celery-beat/deployment.yml
```diff
@@ -0,0 +1,51 @@
...
+          value: {{.Values.database.dbPort}}
+        volumeMounts:
+        - name: django-fileserver-storage
+          mountPath: /data
+      - name: cloud-sql-proxy
+        image: {{.Values.database.cloudsqlimage}}
+        command:
+        - "/cloud_sql_proxy"
+        - "-ip_address_types=PRIVATE"
+        - "--instances={{.Values.database.instanceConnectionName}}=tcp:{{.Values.database.dbPort}}"
```
tychota, line 44: Well done, you found this and this is an headache !

tychota, line 44: [Importance 2/10]

Idealy we should use configmap + env variable

kubernetes/celery-beat/deployment.yml
```diff
@@ -0,0 +1,51 @@
...
+        securityContext:
+          runAsNonRoot: true
```
tychota, line 46: [Importance 3/10]

Idealy we should go further on security, for instance using `readOnlyRootFilesystem: true`.

I can compare later what i do at work and give more precise recomendation

kubernetes/celery-beat/deployment.yml
tychota: In the mean time, some docs https://sysdig.com/wp-content/uploads/2019/01/kubernetes-security-guide.pdf

(chapter 4 and 5)

kubernetes/celery-beat/deployment.yml
tychota: In the mean time some docs ^^
https://sysdig.com/wp-content/uploads/2019/01/kubernetes-security-guide.pdf

chapter 4 and 5

kubernetes/celery-beat/deployment.yml
```diff
@@ -0,0 +1,51 @@
...
+      volumes:
+      - name: django-fileserver-storage
+        persistentVolumeClaim:
+          claimName: fileserver-claim
+      serviceAccountName: {{.Values.serviceAccounts.cloudSqlStorage}}
```
tychota, line 51: EOF

kubernetes/celery-beat/deployment.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: cloud-sql-proxy
+        image: {{.Values.database.cloudsqlimage}}
+        command:
+        - "/cloud_sql_proxy"
+        - "-ip_address_types=PRIVATE"
+        - "--instances={{.Values.database.instanceConnectionName}}=tcp:{{.Values.database.dbPort}}"
+        securityContext:
+          runAsNonRoot: true
+      volumes:
+      - name: django-fileserver-storage
```
tychota, line 48: Nice, well done

kubernetes/django/deployment.yml
```diff
@@ -0,0 +1,69 @@
...
+      labels:
+        app: django
+    spec:
+      containers:
+      - name: django
+        image: {{.Values.django.image.repo}}:{{.Values.django.image.tag}}
+        ports:
+        - name: django-port
+          containerPort: {{.Values.django.containerPort}}
+          protocol: TCP
```
tychota, line 23: nice

kubernetes/celery-beat/deployment.yml
```diff
@@ -0,0 +1,51 @@
...
+  selector:
+    matchLabels:
+      app: celery-beat
+  template:
+    metadata:
+      labels:
+        app: celery-beat
+    spec:
+      containers:
+      - name: celery-beat
```
tychota, line 20: [Importance 2/10]

Ideally we also needs some limits.
https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/

tychota, line 20: [Importance 4/10]

Ideally we should know if the pod is up and restart if not
we can use https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

kubernetes/celery-worker/deployment.yml
```diff
@@ -0,0 +1,49 @@
+apiVersion: apps/v1
```
tychota, line 1: Same than beat

kubernetes/django/deployment.yml
```diff
@@ -0,0 +1,69 @@
+apiVersion: apps/v1
```
tychota, line 1: Same than beat

kubernetes/django/deployment.yml
```diff
@@ -0,0 +1,69 @@
...
+            # POSTGRES_USER and POSTGRES_PASSWORD
+            name: postgres-env-secrets
+        env:
+        - name: POSTGRES_HOST
+          value: {{.Values.database.dbIpAddressWithinContainer}}
+        - name: POSTGRES_DB
+          value: {{.Values.database.dbName}}
+        - name: POSTGRES_PORT
+          value: {{.Values.database.dbPort}}
+        readinessProbe:
```
tychota, line 38: Nice

kubernetes/ingress.yml
```diff
@@ -0,0 +1,22 @@
+apiVersion: networking.k8s.io/v1beta1
```
tychota, line 1: nice

kubernetes/nginx/deployment.yml
```diff
@@ -0,0 +1,45 @@
...
+          periodSeconds: 20
+          timeoutSeconds: 3
+          failureThreshold: 5
+        volumeMounts:
+        - name: nginx-fileserver-storage
+          mountPath: /data
+      volumes:
+      - name: nginx-fileserver-storage
+        persistentVolumeClaim:
+          claimName: fileserver-claim
```
tychota, line 45: EOF

kubernetes/celery-beat/deployment.yml
```diff
@@ -0,0 +1,51 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: celery-beat-deployment
+  labels:
+    app: celery-beat
```
tychota, line 6: [Importance 6/10]

We should dod this trick https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments

kubernetes/celery-beat/deployment.yml
```diff
@@ -0,0 +1,51 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: celery-beat-deployment
+  labels:
```
tychota, line 5: [Importance 8/10]

We are missing labels/annotation used by helm to see what it manage

https://helm.sh/docs/chart_best_practices/labels/

tychota:
You did well, that is a good first step.

---

For it to work you need to respect helm structure.
You can read more here: https://helm.sh/docs/chart_template_guide/getting_started/

For instance it would make sense to have threehelms chart for now.
- one with django and celery beat/worker, and associated service / ingress / secrets / configmap
- one with nginx, and associated service / ingress
- (one with redis (clusterised, maybe), and associated service / ingress)

For redis we may  want to use an operator, like https://github.com/spotahome/redis-operator (there is a spooky  or the one from redis itself https://redislabs.com/blog/redis-enterprise-operator-kubernetes/
We should decide if we want or not: I'm not sure, open to discussion https://medium.com/swlh/production-checklist-for-redis-on-kubernetes-60173d5a5325
If you want you can read more about the benefir of operators https://coreos.com/operators/ and https://kubernetes.io/docs/concepts/extend-kubernetes/operator/ and https://blog.container-solutions.com/kubernetes-operators-explained but the rule of thumb is that stateful (eg database) running in kubernetes needs an operator.

--- 

> Redis is missing from this push, but I think mostly everything else is here aside from secrets?

secrets shouldn't be part of the helm template but created separately

you also miss the "Job" part in django 

https://itnext.io/database-migrations-on-kubernetes-using-helm-hooks-fb80c0d97805

 ---

> the expected workflow for using these files

First we are going to build images and upload them to registry

Then install template in cluster using something like in https://helm.sh/docs/chart_template_guide/getting_started/#a-first-template

```bash
helm install katago-server ./katago-server
```

Then later, after upgrading the version in values.yaml

```bash
helm upgrade katago-server ./katago-server
```

https://helm.sh/docs/intro/using_helm/#helm-upgrade-and-helm-rollback-upgrading-a-release-and-recovering-on-failure

---

> versioning and tagging and upload of images

We can use git destribe for the version

Eg

```
docker build -t grc.io/org/proj/katago-server:$(git describe --abbrev=7 --tags --always --first-parent) -t katago-server:latest ./path/to/katago/dockerfile
docker push grc.io/org/proj/katago-server:$(git describe --abbrev=7 --tags --always --first-parent)
docker push grc.io/org/proj/katago-server:latest
```

---

>  how secrets work, how internal services communication and hostname works for redis and celery

not sure what you mean

Secret:
- are crd (custom resource) for kube, store normally in etcd as clear base64 but in our cluster we use google kms
- https://kubernetes.io/docs/concepts/configuration/secret/#security-properties
- https://cloud.google.com/kubernetes-engine/docs/how-to/encrypting-secrets

Communication kubernetes:
- this is the detailled way kubernetes networking works by default https://sookocheff.com/post/kubernetes/understanding-kubernetes-networking-model/
- the networking itself is an abstraction so you can change the way it works: it use a CNi (and you can even define your own CNI for networking if you are crazy https://www.altoros.com/blog/kubernetes-networking-writing-your-own-simple-cni-plug-in-with-bash/): for instance cilium use e-bpf https://www.youtube.com/watch?v=_Iq1xxNZOAo and calisato use l2 and BGP https://docs.projectcalico.org/networking/determine-best-networking
- google does the hard job for you: by default like normal (veth, bridge, arp layer 2) https://cloud.google.com/kubernetes-engine/docs/concepts/network-overview and since a few month you can opt in cillium https://cloud.google.com/blog/products/containers-kubernetes/bringing-ebpf-and-cilium-to-google-kubernetes-engine
- basically you don't have to care how it is done except you need some really specif stuff

DNS in kubernetes:
- kube setup a dns for you: https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/
- https://cloud.google.com/kubernetes-engine/docs/concepts/service-discovery

kubernetes/celery-beat/deployment.yml
lightvector: Thanks, will fix in upcoming commit.

compose/production/django/entrypoint
```diff
@@ -14,7 +14,7 @@ if [ -z "${POSTGRES_USER}" ]; then
     base_postgres_image_default_user='postgres'
     export POSTGRES_USER="${base_postgres_image_default_user}"
 fi
-export DATABASE_URL="postgres://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:${POSTGRES_PORT}/${POSTGRES_DB}"
+export DATABASE_URL="${POSTGRES_HOST}://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:${POSTGRES_PORT}/${POSTGRES_DB}"
```
tychota, line 17: revert to postgres://

kubernetes/celery-beat/deployment.yml
tychota: I can look at it ?

kubernetes/celery-beat/deployment.yml
lightvector: I did a partial step - I moved everything out of default namespace but I didn't go with the complication of having different namespaces for different groups of pods.

kubernetes/celery-beat/deployment.yml
lightvector: Commented why in the file as discussed, thanks.

kubernetes/celery-beat/deployment.yml
lightvector: Current thing is fine, don't want to bother reworking images. :)

kubernetes/celery-beat/deployment.yml
lightvector: The links we looked at only works for celeryworkers, it doesn't serve as a check for celerybeat. (I tested this out on the test prod server manually killing or suspending celerybeat and seeing what happened).

So I added the liveness probe for the worker, but for celerybeat for now I give up and just let it fail if it fails.

compose/production/django/entrypoint
lightvector: done

kubernetes/celery-beat/deployment.yml
lightvector: done

kubernetes/celery-beat/deployment.yml
lightvector: I did it, thanks.

kubernetes/celery-beat/deployment.yml
lightvector: Done

kubernetes/celery-beat/deployment.yml
lightvector: I think this is done now, thanks.

kubernetes/katago-server/templates/celery-flower/deployment.yaml
```diff
@@ -0,0 +1,79 @@
...
+    matchLabels:
+      app: celery-flower
+      {{- include "katago-server.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      labels:
+        app: celery-flower
+        {{- include "katago-server.selectorLabels" . | nindent 8 }}
+      annotations:
+        checksum/django_env_secrets: {{ include (print $.Template.BasePath "/env/django_env_secrets.yaml") . | sha256sum }}
```
tychota, line 22: How does that work ? does it require to modify the template ?

kubernetes/katago-server/templates/celery-flower/deployment.yaml
```diff
@@ -0,0 +1,79 @@
...
+          tcpSocket:
+            port: celery-flower-port
+          initialDelaySeconds: 15
+          periodSeconds: 20
+          timeoutSeconds: 3
+          failureThreshold: 5
+        resources:
+          limits:
+            cpu: 100m
+        volumeMounts:
```
tychota, line 60: Shall we mount the main storage ?
I guess we can juste use a small amount of cloud storage

Eg
https://cloud.google.com/kubernetes-engine/docs/how-to/stateful-apps?hl=fr#creating_a_statefulset

kubernetes/katago-server/templates/celery-flower/deployment.yaml
```diff
@@ -0,0 +1,79 @@
...
+        - name: django-fileserver-storage
+          mountPath: /data
+          subPath: {{.Values.fileserver.subPath}}
+      - name: cloud-sql-proxy
```
tychota, line 64: Also not sure we need that. 

Flower should not connect to DB

kubernetes/katago-server/values.yaml
```diff
@@ -19,6 +19,10 @@ celeryWorker:
   replicaCount: 1
 celeryBeat:
   replicaCount: 1
+celeryFlower:
```
tychota, line 22: nice

kubernetes/katago-server/templates/ingress.yaml
```diff
@@ -22,4 +22,8 @@ spec:
       - path: /
         backend:
           serviceName: django-service
-          servicePort: {{.Values.django.servicePort}}
\ No newline at end of file
+          servicePort: {{.Values.django.servicePort}}
+      - path: /
```
tychota, line 26: That looks false to me. You can't have two same path for the same host.

I will create another host eg `flower.katagotraining.org`

kubernetes/katago-server/templates/django/migrate.yaml
```diff
@@ -0,0 +1,49 @@
+apiVersion: batch/v1
+kind: Job
```
tychota, line 2: nice

kubernetes/katago-server/templates/django/migrate.yaml
```diff
@@ -0,0 +1,49 @@
...
+metadata:
+  name: django-migrate-job
+  labels:
+    {{- include "katago-server.labels" . | nindent 4 }}
+spec:
+  backoffLimit: 2
+  ttlSecondsAfterFinished: 900
```
tychota, line 14: Why such an hight number of seconds ? Maybe it is ok but i don't understand the use case

kubernetes/katago-server/templates/django/migrate.yaml
```diff
@@ -0,0 +1,49 @@
...
+        - configMapRef:
+            # GCP_DATABASE_INSTANCE and POSTGRES_PORT
+            name: postgres-env-config
+        securityContext:
+          runAsNonRoot: true
+      volumes:
+      - name: django-fileserver-storage
+        persistentVolumeClaim:
+          claimName: fileserver-claim
+      serviceAccountName: {{.Values.serviceAccounts.cloudSqlStorage}}
```
tychota: EOF

kubernetes/katago-server/templates/celery-beat/deployment.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
   namespace: serverspace
   labels:
     app: celery-beat
+    app.kubernetes.io/component: celery-beat
```
tychota, line 8: nice

tychota (approved):

tychota:
image

kubernetes/katago-server/templates/django/migrate.yaml
lightvector: I want to be able to see the container for a while after the job inside it failed, to make it easier to inspect it and look at logs and so on. :)

kubernetes/katago-server/templates/ingress.yaml
lightvector: You can't have two paths on the same host even if they are different ports?
Flower is on port 5555, right? This is how we had it in docker-compose production.yml too, right?. We have path "/" with port 80 or port 443 go to django, except that if also has the prefix "/media" it goes nginx. And we had path "/" on port 5555 to to flower?

kubernetes/katago-server/templates/celery-flower/deployment.yaml
lightvector: I am simply doing the recommendation that *you* linked! Right?
https://github.com/katago/katago-server/pull/44#discussion_r517000159

My understanding of workflow was that we will have a secrets file env/django_env_secrets.yaml sitting on disk (unencrypted while it is on disk?) that we deploy, the same way we have a secrets file in docker compose sitting (unencrypted) on disk that is mounted by production.yml into the environment variables for a container.

This uses the mechanism in the link you have - if the secrets file changes, then the sha256sum will change, and so this annotation will change, and then helm will know that this deployment needs to be updated and to do a rolling restart properly.

lightvector: Is there a different workflow you expected?

kubernetes/katago-server/templates/celery-flower/deployment.yaml
lightvector: In production.yml we simply used exactly mirrored settings for all of the celery containers, compared to the django container. Same mounts, same environment variables, etc. I'm simply doing that here, since I wasn't sure what would be safe to be different or not. There should be no harm, right?

kubernetes/katago-server/templates/celery-flower/deployment.yaml
lightvector: In production.yml we simply used exactly mirrored settings for all of the celery containers, compared to the django container. Same mounts, same environment variables, etc. I'm simply doing that here, since I wasn't sure what would be safe to be different or not. There should be no harm, right?

lightvector: But okay, if you are sure flower doesn't need it, I can remove it. :)

tychota: Normally it shouldn't, if it does it is a bug for me. Meaning I'm not actually sure it doesn't need.

kubernetes/katago-server/templates/celery-flower/deployment.yaml
tychota: makes sense.

I wasn't planing to deploy secrets with helm so I was thinking of only doing checksum for configmap (that can be deploy with helm, same as https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments).

Also based on

image

I'm not sure we need it for env variables (sorry if I made you do useless stuff).

kubernetes/katago-server/templates/celery-flower/deployment.yaml
tychota: Except if there is a bug in flower and someone can delete all file from the flower container.

Ideally I would try in production to just give what container needs.

kubernetes/katago-server/templates/ingress.yaml
tychota: I'm almost certain it won't work: https://cloud.google.com/kubernetes-engine/docs/concepts/ingress

The ingress itself is a loadblancer listenning on only 80 and 443 (like traefik by default)

image

to be able to reach `whatever.url.com:5555` we do not only need the internal pod to listen on 5555 but also the ingress.
On traefik we opened a new endpoint for this.

On GCP i don't know how to do, with defult ingress controller. (with another ingress controller, you can do this for instance https://kubernetes.github.io/ingress-nginx/user-guide/exposing-tcp-udp-services/)

My point isnon standard port like 5555 is non standard so we should prefer having two hosts

kubernetes/katago-server/templates/django/migrate.yaml
tychota: ok for me

kubernetes/katago-server/templates/ingress.yaml
tychota: Also we have not configured https yet: here is how to do https://cloud.google.com/kubernetes-engine/docs/how-to/managed-certs

lightvector: Maybe we should just use nginx ingress controller then?

kubernetes/katago-server/templates/celery-flower/deployment.yaml
lightvector: Okay so technically in the docker compose config we should also change it to not mount the database too, right?

kubernetes/katago-server/templates/celery-flower/deployment.yaml
lightvector: Okay so technically in the docker compose config we should also change it to not mount the database and the storage too, right?

kubernetes/katago-server/templates/celery-flower/deployment.yaml
tychota: I think yes, but not sure if it works

kubernetes/katago-server/templates/celery-flower/deployment.yaml
tychota: I think yes, but not sure if it works

kubernetes/katago-server/templates/ingress.yaml
tychota: not sure we really need to expose a custom port

we can have

`katagotraining.com -> django`
`flower.katagotraining.com ->flower`

that would be more clean for sure

kubernetes/katago-server/templates/frontendconfig.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: networking.gke.io/v1beta1
+kind: FrontendConfig
+metadata:
+  name: katago-server-frontend-config
+spec:
+  redirectToHttps:
+    enabled: true
+    responseCodeName: MOVED_PERMANENTLY_DEFAULT
```
tychota, line 8: EOF

PR #45, opened 2020-11-06
Remove datadog

lightvector:
Entirely removes datadog, in preparation for Kubernetes where we won't have set up datadog.


lightvector:
Also, fixes a bug where gunicorn was hardcoded to use 9 workers, regardless of WEB_CONCURRENCY setting.

compose/production/django/start
```diff
@@ -6,4 +6,4 @@ set -o nounset
 
 
 python /app/manage.py collectstatic --noinput
-ddtrace-run /usr/local/bin/gunicorn src.asgi:application --bind 0.0.0.0:5000 --chdir=/app --timeout 300 --workers=9 -k uvicorn.workers.UvicornWorker
+/usr/local/bin/gunicorn src.asgi:application --bind 0.0.0.0:5000 --chdir=/app --timeout 300 --workers="${WEB_CONCURRENCY}" -k uvicorn.workers.UvicornWorker
```
tychota, line 9: nice.

Also we may want to get rid of gunicorn too since we will use kubernetes
https://github.com/tiangolo/fastapi/issues/1484

https://www.uvicorn.org/#command-line-options

tychota (approved):

compose/production/django/start
lightvector: Let's keep gunicorn for now. I was just planning to just set the number of replicas to 2 or 3, for the redundancy, and so I'm okay with there being a gunicorn doing some multiprocessing inside each container to also be able to handle multiple requests.

If it turns out for some reason we need lots more pods, we'd probably want to increase the limit for mumber of machines or make the machines each have more cores and memory, which would be some manual work anyways.

PR #46, opened 2020-11-07
Kubernetes fixes

lightvector:
I was starting to write the values file and when I got to the part for nginx I ran into this. So this pull request fixes nginx config and also a few other annotations.


lightvector:
Going ahead to merge this so that we can make a proper new tag. There will of course still be some more changes and fixes we need later.

PR #50, opened 2020-11-11
Improve users and build

lightvector:
* We now run nginx as non-root, and disable privilege escalation on all containers.
* We fix django to use a hardcoded high-id non-system user and group, so that it has no special permissions and does not conflict with any user ids outside the container, especially when it writes to the filesystem.
* We backport a few minor dockerfile optimizations from the latest django cookiecutter.
* All containers are running as non root now, despite `runAsNonRoot` not working (it doesn't work in kubernetes because kubernetes doesn't support checking users by username rather than by integer).
* We add a bunch of useful scripts to the repo that I've been using, for convenience.

### DIFF
--- a/compose/production/nginx/default.conf
+++ b/compose/production/nginx/default.conf
@@ -1,5 +1,5 @@
 server {
-   listen                      80; # listen on http
+   listen                      18080; # listen on http
    server_name                 _; # don't expect a special server name (it is traefik job to already do the filter)
    client_max_body_size        1000M; # We may increase that for network
    set                         $cache_uri $request_uri;
--- a/compose/production/nginx/nginx.conf
+++ b/compose/production/nginx/nginx.conf
@@ -1,0 +1,31 @@
+# user  nginx;  # disabled to avoid a warning since we're booting up the container as nginx, rather than as root
+worker_processes  1;
+
+error_log  /var/log/nginx/error.log warn;
+pid        /var/run/nginx.pid;
+
+
+events {
+    worker_connections  1024;
+}
+
+
+http {
+    include       /etc/nginx/mime.types;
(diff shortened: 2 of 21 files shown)


compose/production/nginx/default.conf
```diff
@@ -1,5 +1,5 @@
 server {
-   listen                      80; # listen on http
+   listen                      18080; # listen on http
```
tychota, line 2: standard port when you can't 80 is 8080

compose/production/nginx/nginx.conf
```diff
@@ -0,0 +1,31 @@
...
+                      '"$http_user_agent" "$http_x_forwarded_for"';
+
+    access_log  /var/log/nginx/access.log  main;
+
+    sendfile        on;
+    #tcp_nopush     on;
+
+    keepalive_timeout  65;
+
+    #gzip  on;
```
tychota, line 28: Why off ?

tychota (approved):

compose/production/nginx/default.conf
lightvector: I know, but I don't see why it matters. :)

compose/production/nginx/nginx.conf
lightvector: This was what the config was by default. The default official nginx image has this line commented out.

tychota: Maybe we can activate it to save some bandwidth

lightvector: Maybe, but bandwidth for nginx should be completely trivial - the only hits to the nginx will be when people access individual SGF files manually in the browser, because networks downloads are hosted on cloud bucket.

PR #51, opened 2020-11-12
Improve networking and limits

lightvector:
* Sets a large value for strict transport security now.
* Fixes some missing fields in the example files and quotes the strings the way the actual real production values_private and other files needed to do.
* Adds limits and requests to help kubernetes decide how to scale the cluster and arrange pods
* Adds topology constraints. However, they don't seem to be working - the cluster seems to be choosing a distribution that violates them for some reason. Not sure why.


lightvector:
Added a whole ton of work today - got real IP address tracing to work in both django and nginx despite the layers of Google cloud proxies in between messing up the ip address. Then, used to add rate limits, and I tested that both the django and nginx rate limit were working. Nginx build is cleaned up and overhauled to use configmap instead of custom volume mounts, and to specify things based on environment variables.

Everything is tested and working, so I merge this, but of course if you decide to review it and see any problems, let me know.

PR #54, opened 2020-11-22
Optimize docker image

tychota:


compose/production/django/Dockerfile
```diff
@@ -70,5 +75,6 @@ COPY --chown=django:django src  /app/src
 COPY --chown=django:django manage.py  /app
 
 USER django
+EXPOSE 8000
```
lightvector, line 78: Could you explain what this is for? Why do we need it now and why did we not need it before?

kubernetes/build_images_for_gcp.sh
```diff
@@ -3,13 +3,13 @@
 IMAGEREPOPATH=$(cat imagerepopath.txt | tr -d '\n')
 IMAGETAG=$(git describe --abbrev=7 --tags --always --first-parent)
 
-docker build \
-       -t "$IMAGEREPOPATH"/katago-django:"$IMAGETAG" \
-       -t "$IMAGEREPOPATH"/katago-training-server/katago-django:latest \
```
lightvector, line 8: Good catch.

compose/local/postgres/postgres.conf
```diff
@@ -0,0 +1,750 @@
+# -----------------------------
```
lightvector, line 1: This config was not here before so I presume it there was a default that was part of the image that we were using before.

Why did we need to add this config, and are there any parameters in this config are now set differently than the default that was set before?

compose/local/postgres/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM postgres:12.3
```
lightvector, line 1: Could you explain a bit more about the maintenance and backup scripts that were added here? And is there anything notable that needs to happen in order to start using this image?

For example, on the ongoing test run in test.katagodistributed.org, is it still sufficient to simply build, migrate, and restart the server - will the new postgres image still connect fine to the old database volume and keep working without any special changes? Thanks!

compose/local/django/Dockerfile
```diff
@@ -42,5 +48,6 @@ RUN chmod +x /start-flower
 RUN mkdir /data
 
 WORKDIR /app
+EXPOSE 8000
```
lightvector, line 51: What is this for and why is it needed now when it was not needed before?

compose/local/postgres/Dockerfile
lightvector: Oh wait, this is all local. Is there a reason for doing all this work for "local" but not add the same maintenance and backup scripts for "production"?

lightvector: Ooooh. We did have this config in production. Okay. Is the reason for copying it to local simply to make it more consistent, or is there a direct use case in local right now that you needed to fix a problem when changing the image?

tychota: We were using the production postgres inside local.yml.
Which is weird, for me.

tychota: I didn't change anything but it is normal to have different query in postgres for dev and prod.

tychota: Eg: http://www.louisemeta.com/blog/developers-and-logs/

compose/production/django/Dockerfile
tychota: It is indeed a leftover (but no harm and probably more standard).

image

I added it so docker-slim could work (docker-slim look at the EXPOSE port and probe the api).

compose/local/postgres/postgres.conf
tychota: We were using the production postgres inside local.yml.
Which is weird, for me.

lightvector:
I pushed one change to revert the need to specify python3 everywhere and instead just make python3 the default. 
And fixed a bug where port 8000 was exposed instead of 5000. Look good otherwise.

PR #56, opened 2020-12-05
Lots of fixes

lightvector:
Note: these changes have all been tested - they are being used live on test.katagodistributed.org with no problems, and most of them have also been used during my uploading of startposes and pregeneration of rating games on katagotraining.org.

Here is a list of the changes:
* Add ability to select rating games networks based on low data. I found that the current setting of highest uncertainty wasn't the best, because sometimes the most uncertain network might be one whose Elo is far away from others, but it is hard to reduce and each game contributes very little. This may cause other networks to starve and never get many ratings games, for a very long time. In that case it is better to also include sometimes games with networks that simply have few games.

* Add a direct rating game count to the network model as a record of the exact status of games that the Elo calculations were based on, and display it in the networks view.

* Fixes a bug where some counts of total rating games on the home page were doubled (it counted the game for both black AND white, which of course means you double-count the game).

* Shannon entropy was crashing on the kubernetes site due to numeric instability with large Elo differences, by returning NAN and bad things happening. Unfortunately, due to the nesting of logarithms, it is very hard to fix and write stably. Therefore, we switch to a slightly simpler formula and instead distribute games based on the variance of the bernoulli win/loss result. This is *extremely* similar to Shannon entropy, but is easy to write in a numerically stable way.

* wgo.js was mistakenly configured to treat self-capture ("suicide") moves as illegal in game records. Some of KataGo's games have it as a legal move though, so we enable it so we don't cause errors when wgo.js attempts to view such a game in the browser game replay.

* Ratings networks selections were heavily biased because when ratings were *exactly* equal, it returned 100% of the time based on a fixed arbitrary postgres-specific ordering, presumably whatever postgres's index happened to have. You might think this is not a case to worry about, but exactly equal ratings do happen in real usage if we upload more than one network at once - the new networks may be *exactly* equal because there are no games to tell the difference between them! And this is a big problem when you upload 50+ new networks at the same time to "start" a run, like I just did a few days ago. So we modify our ratings games network queries to break ties randomly now, by including random() in the order_by condition.

* Improved formatting for some games lists - deleted useless columns, colored the results based on win/draw/loss.

* Elo graph legend was overflowing the boundary, so we adjust the legend placement. Also we now view log of data rows by default.

* The site was sometimes crashing when there was very little data on the site due to counts from database materialized views evaluating to None. Fixed now.

* Email admin page is restored, and now there is a page to allow users to change password and update email address. Customizes the email message sent for these things too.

* Email verification is now required for django rest framework. We already required email verification for django site login, but unfortunately django rest framework does NOT pay attention to the same settings.py options that django does. So we manually implement a check.

* Fix a bug where unauthenticated users were allowed to attempt to upload training or ratings games. This would result then in http 500 server errors because the server would attempt to fill in "submitted_by" with a nonexistent username.

* Fix bug where bad things might happen to Elo calculation if somehow a network became its own parent.

* Validation checks in the save() method of a model appear to cause http 500 server internal errors in some cases. We move those instead into the clean() method of the model. Unfortunately, the clean() method is not called by the django rest framework, so we slightly duplicate some logic and also add those checks into the django rest framework serializer fields validations.

* Reliance on python-magic turns out to be a bad idea. Extremely rarely, it is possible for legal zip files to be detected instead as DOS boot sectors. I suspect it may also be possible for similar bad detections to happen to legitimate gzip files and maybe even legitimate sgf files. So, we add custom file validation tests that manually special-case the file types we care about and apply correct tests for them that should never fail to reject valid files. We fall back to python-magic for detection outside of those cases. Also an example of such a zip file has been added to one of the pytests. 

* Persistent volumes are treated immutable in Kubernetes by default. So we include a version in the name so that when we want to change it, we can make Helm push newly-named volumes that won't cause conflicts and problems. This has been used successfully to switch katagotraining.org from Standard to SSD filestore, so it works now with this fix.

* Adds a bunch of pytest tests for various new things relating to the above changes.


tychota:
> Note: these changes have all been tested - they are being used live on test.katagodistributed.org with no problems, and most of them have also been used during my uploading of startposes and pregeneration of rating games on katagotraining.org.

Nice and thanks for the high level list of changes.

src/apps/trainings/models/network.py
```diff
@@ -128,6 +128,7 @@ class Meta:
     log_gamma_upper_confidence = FloatField(
         _("log gamma upper confidence"), default=0, db_index=True, help_text=_("Upper confidence bound on BayesElo strength of network."),
     )
+    log_gamma_game_count = BigIntegerField(_("log gamma game count"), default=0, db_index=True, help_text=_("Number of real games used to compute log_gamma for this network."),)
```
tychota, line 132: You prefixed it with log_gamma because it can be different from the actual game count ? (eg new game came but elo wasn't updated ?)

src/apps/trainings/services/bayesian_elo.py
```diff
@@ -54,17 +54,23 @@ def update_ratings_iteratively(self, number_of_iterations):
                 self._update_specific_network_log_gamma(network_id, network_log_gamma)
             self._reset_anchor_log_gamma()
 
+        # Count games by network
+        df = pandas.DataFrame()
+        df["reference_network"] = self._detailed_tournament_results["reference_network"]
+        df["nb_games"] = 0.0
+        df["nb_games"] += self._detailed_tournament_results["total_games_white"]
+        df["nb_games"] += self._detailed_tournament_results["total_games_black"]
```
tychota, line 62: I'm rusty with panda but is `new_df["col1"] = old_df["col1"]` keeping the index of old_df (so that network are sumed, even if the order differs ?

Or does it really on having everything in the same order ?

src/apps/distributed_efforts/services/rating_network_pairer.py
```diff
@@ -101,7 +102,7 @@ def _choose_opponent(self, reference_network):
 
         # Vary the reference net's log gamma so that networks that are uncertain will play a greater variety of opponents
         # based on that uncertainty
-        ref_net_log_gamma = reference_network.log_gamma + np.random.normal() * reference_network.log_gamma_uncertainty
+        ref_net_log_gamma = reference_network.log_gamma + np.random.normal() * reference_network.log_gamma_uncertainty * self.current_run.rating_game_variability_scale
```
tychota, line 105: This scale is supposed to increase the "deviation" where network can be chosen ?

src/apps/distributed_efforts/services/rating_network_pairer.py
```diff
@@ -130,9 +131,25 @@ def _choose_opponent(self, reference_network):
...
-        ])
-        shannon_entropy = -win_probability * np.log2(win_probability) - (1 - win_probability) * np.log2(1 - win_probability)
+        log_variances = [
+            self._log_variance_of_gamma_difference((opp_net.log_gamma - ref_net_log_gamma) / self.current_run.rating_game_variability_scale) for opp_net in nearby_networks
+        ]
+        max_log_variance = max(log_variances)
+        # Subtract out the max to make sure that we're near 0, for numerical stability
+        log_variances = [ log_variance - max_log_variance for log_variance in log_variances ]
+        variances = [ exp(log_variance) for log_variance in log_variances ]
+        print(log_variances)
```
tychota: Maybe remove the print.

ideally use logging in debug (we need more logs on the server)

src/apps/distributed_efforts/services/rating_network_pairer.py
```diff
@@ -130,9 +131,25 @@ def _choose_opponent(self, reference_network):
...
+        print(variances)
+        return np.random.choice(nearby_networks, p=variances / np.sum(variances))
```
tychota, line 143: Not sure I understand the math, can you explain me on discord how it work ? (just curious)

src/apps/distributed_efforts/services/rating_network_pairer.py
tychota: Answered in migration

src/frontend/templates/pages/sgfplayer.html
```diff
@@ -45,6 +45,7 @@
     sgfFile: "{{ game.sgf_file.url }}",
     board: { background: "#E8CF6D" },
     layout: dynamicLayout,
+    allowIllegalMoves: true,
```
tychota, line 48: :)

src/apps/trainings/managers/network_queryset.py
```diff
@@ -39,25 +39,25 @@ def select_most_recent(self, run: Run, for_training_games=False, for_rating_game
 
     def select_high_lower_confidence(self, run: Run, for_training_games=False, for_rating_games=False):
         filtered = self.select_networks_for_run(run,for_training_games=for_training_games,for_rating_games=for_rating_games)
-        return filtered.order_by("-log_gamma_lower_confidence").first()
+        return filtered.order_by("-log_gamma_lower_confidence","?").first()
```
tychota, line 42: what is the impact of the random here ? Won't it annihilate every index optimization if the table is big enough ?

src/apps/distributed_efforts/tests/test_api.py
```diff
@@ -463,7 +463,7 @@ def test_high_elo(self):
         s = ""
         for k,v in sorted(counts.items()):
             s += k + ":" + str(v) + ", "
```
tychota: Maybe we can extract this into an utils

src/apps/games/models/rating_game.py
```diff
@@ -20,3 +20,9 @@ class Meta:
         verbose_name = _("Rating game")
         ordering = ["-created_at"]
 
+    def save(self, *args, **kwargs):
```
tychota: nice

src/apps/trainings/models/network.py
lightvector: Yes, this specifically is the game count that is matching with the Elo computed.

src/frontend/templates/pages/games_list_by_network.html
```diff
@@ -22,15 +22,52 @@ <h1 class="title">{% blocktrans with network_name=network.name %} Rating Games f
...
+    <td> White </td>
+    <td> {{ game.black_network.name|chop_network_run_name:run.name }} </td>
     {% endif %}
+    <td> {{ game.board_size_x }} </td>
+    <td> {{ game.board_size_y }} </td>
+    <td> {{ game.komi }} </td>
+    <td> {{ game.gametype|underscores_to_spaces }} </td>
+    <td> {{ game.game_length }} </td>
+    <td> {{ game.result_text }} </td>
+    <td> <a href="{{ game.sgf_file.url }}">[{% trans "SGF" %}]</a> </td>
```
tychota, line 58: We should update the po/mo https://github.com/katago/katago-server/tree/master/locale

I can translate in French

src/apps/distributed_efforts/services/rating_network_pairer.py
lightvector: Oops hehe good catch on the print.

src/frontend/static/css/custom.css
```diff
@@ -0,0 +1,13 @@
+
+.winnerResultStyle {
+  color:#00cc77;
+  font-weight:500;
+}
+
+.loserResultStyle {
+  color:red
```
tychota, line 8: Maybe we can reuse bulma red https://bulma.io/documentation/helpers/color-helpers/

src/apps/distributed_efforts/services/rating_network_pairer.py
lightvector: Wait, I did at least comment out the print lat in a later commit I think. Seems okay now.

src/apps/trainings/services/bayesian_elo.py
lightvector: I'm not entirely sure what you're asking. Can you explain what specifically you are worried about with the current code that might have a bug? Which of the indices are you worried about a mismatch here in this Elo game count computation?

src/apps/distributed_efforts/services/rating_network_pairer.py
lightvector: Ok.

src/frontend/templates/pages/games_list_by_network.html
lightvector: What is a po/mo?

src/frontend/static/css/custom.css
lightvector: Maybe, but I think it is clean to have a class that corresponds semantically to "here is how to format the losing result" instead of simply hardcoding a color class?

src/frontend/templates/pages/account.html
```diff
@@ -0,0 +1,28 @@
...
+
+{% block title %}{% trans "KataGo - Account" %}{% endblock %}
+
+
+{% block content %}
+
+<h2 class="title">{% trans "KataGo - Account" %}</h1>
+
+{% url 'account_change_password' as password_url %}
+<a href="{{password_url}}">
```
tychota, line 15: Does the password_url exist ?

src/frontend/templates/pages/account.html
```diff
@@ -0,0 +1,28 @@
...
+  <button class="button is-warning">
+    {% trans "Change password" %}
+  </button>
+</a>
+
+{% url 'account_email' as email_url %}
```
tychota, line 21: Does the account_email exist ?

src/frontend/templates/pages/account.html
```diff
@@ -0,0 +1,28 @@
...
+<a href="{{email_url}}">
+  <button class="button is-warning">
+    {% trans "Change email" %}
+  </button>
+</a>
+
+{% endblock content %}
```
tychota, line 28: I would expect some forms, maybe they are already here

src/urls.py
```diff
@@ -50,6 +51,7 @@
         # Django Admin, use {% url 'admin:index' %}
         path(settings.ADMIN_URL, admin.site.urls),
         path(f"{settings.ADMIN_URL}doc/", include("django.contrib.admindocs.urls")),
+        path("account/", AccountView.as_view(), name="account_page"),
```
tychota, line 54: It is a bit weird (not sure myself what best) to have account/ and accounts/

src/apps/games/models/rating_game.py
```diff
@@ -20,9 +21,10 @@ class Meta:
...
         ordering = ["-created_at"]
 
-    def save(self, *args, **kwargs):
+    def clean(self):
         # Ratings games must involve distinct networks
         if self.white_network == self.black_network:
-            raise ValueError("Ratings games cannot be between a network and itself")
+            raise ValidationError("Ratings games cannot be between a network and itself")
 
+    def save(self, *args, **kwargs):
```
tychota, line 29: Maybe we can remove save override.

src/apps/games/serializers/rating_game.py
```diff
@@ -49,6 +50,8 @@ class Meta:
         }
 
     def validate(self,data):
+        if data["white_network"] == data["black_network"]:
```
tychota, line 53: Is it mandatory/good to do this logic twice ?

src/apps/trainings/models/network.py
```diff
@@ -149,10 +150,11 @@ def elo_uncertainty(self):
...
 
-    def save(self, *args, **kwargs):
+    def clean(self):
         # Allow blank file only if random
         no_model_file = not self.model_file or len(self.model_file) <= 0
         if no_model_file and not self.is_random:
-            raise ValueError("model_file is only allowed to be blank when is_random is True")
+            raise ValidationError("model_file is only allowed to be blank when is_random is True")
 
+    def save(self, *args, **kwargs):
```
tychota, line 159: Same, maybe we can remove

src/frontend/templates/account/email/base_message.txt
```diff
@@ -0,0 +1,7 @@
+{% load i18n %}{% autoescape off %}{% blocktrans with site_name=current_site.name %}Hello from {{ site_name }}!{% endblocktrans %}
```
tychota, line 1: Shouldn't we do html email ?

You can use https://mjml.io/try-it-live/templates/recast and edit on the left, on the right you will see the email and an html that is well supported on mail platforms.

src/apps/games/tests/bad_libmagic_npz.py
```diff
@@ -0,0 +1,174 @@
+# This particular npz file is valid but libmagic detects it as a DOS boot sector instead of a zip file
+base64data=b"""UEsDBBQAAgAIABcHhFHviuZpaAcAAMC+AAAVAAAAYmluYXJ5SW5wdXROQ0hXUGFja2Vk7d3Pb9xE
```
tychota, line 2: Weird, why that ?

https://github.com/threatstack/libmagic/blob/1249b5cd02c3b6fb9b917d16c76bc76c862932b6/magic/Magdir/filesystems#L542 ?

src/apps/trainings/services/bayesian_elo.py
```diff
@@ -118,7 +118,8 @@ def _add_virtual_draws(self):
 
             # Everything blows up if for some reason (eg first network, or deleted network)
             # the parent_network_id does not reference an actual network, so let's check that
-            if parent_network_id in network_ids:
+            # Also if somehow a network is its own parent, ignore that for computing ratings
+            if parent_network_id in network_ids and parent_network_id != network_id:
```
tychota, line 122: nice

tychota (approved):
Really nice, thank you for all those fixes :)

src/frontend/templates/pages/account.html
lightvector: Yes, it is defined by django allauth:
https://django-allauth.readthedocs.io/en/latest/views.html

src/frontend/templates/pages/account.html
lightvector: Yes, it is defined by django allauth:
https://django-allauth.readthedocs.io/en/latest/views.html

src/frontend/templates/pages/account.html
lightvector: They are defined by django allauth:
https://django-allauth.readthedocs.io/en/latest/views.html
So all we need to do on this html page is to link to them, which is what the two buttons are doing.

src/urls.py
lightvector: I want to make absolutely sure that I don't collide with django_allauth's url namespace without having to read docs carefully, so yeah, that's why they're separate. :)

src/apps/games/serializers/rating_game.py
lightvector: Yes. There are some code paths within django itself (like admin page) that run clean() but do not run validate(). And django rest framework runs validate() but not clean().
And annoying is that the type is different in each case. In clean() you have a django model object instance. And in django rest framework, you have a dictionary of raw user post data.

src/frontend/templates/account/email/base_message.txt
lightvector: I'm not going to consider this an important change to make and test. All I did with this base message is to use django_allauth's template and slightly adjust the message - other than a slightly customized wording, this is exactly the default template from django_allauth.

src/apps/games/tests/bad_libmagic_npz.py
lightvector: Because libmagic rules are conflicting and sometimes buggy. If you search online you can find other reports of different cases for different kinds of files where libmagic gets it wrong.

It seems very very hard to always get it right because if you have 1000s of different kinds of files, and each of them has different magic byte rules, well, sometimes those rules will be ambiguous and come into conflict.

lightvector: So that's why we specialize for the specific 3 file types we care about. Since we only deal with 3, we can be confident that our custom magic byte checks will never have false negatives or conflict.

src/apps/distributed_efforts/tests/test_api.py
lightvector: Meh. :)

src/apps/trainings/managers/network_queryset.py
lightvector: Replied in discord chat. Apparently this is a bit complicated, postgres 13 has an "incremental sort" feature that allows it to avoid computing random() on the whole table because random() is only the secondary sort field, not the primary ordering field. So it can sort by the primary field log_gamma_lower_confidence first, and then only use random to break ties after that.

However, it seems like based on the analysis, due to the filter criteria, it seems to be doing something involving the whole table anyways? I don't quite understand why. At least, the random isn't making it worse.

src/apps/trainings/models/network.py
lightvector: Done.

src/apps/games/models/rating_game.py
lightvector: Done.

lightvector:
Pushed another fix where it seems local.yml was not using the postgres.conf properly. And removed redundant save overrides mentioned in above comments, thanks!

src/frontend/templates/pages/games_list_by_network.html
tychota: https://docs.djangoproject.com/fr/3.1/topics/i18n/translation/#localization-how-to-create-language-files

src/frontend/static/css/custom.css
tychota: More "red" itself is 100%, not matching with rest of the site.

```css
loserResultStyle {
  color: hsl(348, 100%, 61%);
}
```

tychota: Same for green

src/frontend/templates/pages/account.html
tychota: Ah ok :)

src/apps/games/serializers/rating_game.py
tychota: Ok, weird. I guess it is fine for now, maybe later we can optimize that.

src/apps/games/tests/bad_libmagic_npz.py
tychota: Then maybe we can remove libmagic alltogether ?

tychota: And simplify docker image to remove libmagic

src/urls.py
tychota: I would say then having "user_settings" or something like this is better.

src/frontend/templates/account/email/base_message.txt
tychota: Fine, later on you (or me) might consider adding html version too https://django-allauth.readthedocs.io/en/latest/advanced.html#sending-email

src/apps/games/tests/bad_libmagic_npz.py
lightvector: It's still useful for generating error messages when an invalid file or wrong format happens - it allows the error message to say the type of the file when it doesn't match, and is still more than 99.9% perfect accurate. 99.9% accurate is not good enough for validation, but it is good enough for error messages, so I think it's okay to keep.

PR #57, opened 2020-12-10
User whitelist and stuff

lightvector:
* Add option to restrict a run to a private user whitelist.
* Require runs to be active to upload games to them via rest framework.
* Add tests for these things.


lightvector:
Also improves an error message to be more helpful.

PR #59, opened 2020-12-13
Make email verification optional for the website part of the site

lightvector:


lightvector:
This is so that users on the website can log in and edit their email without having to verify it (such as if they mistyped email).

Contributing via django rest framework still requires email verification.

PR #67, opened 2020-12-20
Fix unique integrity error

lightvector:
I think due to race conditions, I think it is possible that the database is rejecting a unique key violation, without any earlier validation layer catching the problem. This happened today and was reported on Sentry.

I believe this check for the django rest framework exception handler should fix it and cause it not to become an unhandled exception.


lightvector:
This error message is also crafted to be something that KataGo client will recognize and not retry.

PR #71, opened 2020-12-24
More UI stuff

lightvector:
Add a bit of help text about what the +/- means in the network ratings table.
Add a button to view only the most recent part of the graph.
Add a link to KataGo's formal rules.
Disable the scoring functionality in the sgf player that we are using, because it is misleading and incorrect for KataGo's rules.


src/frontend/templates/run_stats_head.html
```diff
@@ -123,7 +123,7 @@
       var maxDataRowsOr10k = Math.max(maxDataRows,10000);
       var buffer = 0.2 * (maxDataRows - minDataRows);
       buffer += 10000; //10000 data rows
```
tychota, line 125: Maybe we can extract a variable with "10000" as we use that magic value everywhere

tychota (approved):

src/frontend/templates/run_stats_head.html
lightvector: Done, thanks.

tychota (approved):

PR #76, opened 2021-01-18
Fix some frontend bugs involving urls and add tests

lightvector:
* Fixes 500 server error when passing a string for the game id field for sgfplayer view. This is because get_object_or_404 doesn't handle the integer parsing error when the input is not an integer.
* Fixes server error when there is a run with no networks that happens when trying to report the time of the latest network.
* Fixes the fact that networks from the wrong run can be specified in the path for looking at network's games.

Adds some new tests to catch these cases.

### DIFF
--- a/src/frontend/tests/test_basic_urls.py
+++ b/src/frontend/tests/test_basic_urls.py
@@ -1,0 +1,195 @@
+import pytest
+import base64
+
+from django.test import Client
+from django.contrib.auth import get_user_model
+from django.core.files.uploadedfile import SimpleUploadedFile
+
+from src.apps.runs.models import Run
+from src.apps.trainings.models import Network
+from src.apps.games.models import RatingGame, TrainingGame, GameCountByNetwork, GameCountByUser
+from src.apps.games.tests.test_game import goodnpzbase64
+
+pytestmark = pytest.mark.django_db
+
+User = get_user_model()
+
+fake_sha256 = "12341234abcdabcd56785678abcdabcd12341234abcdabcd56785678abcdabcd"
+
+class TestUrlsEmptySite:
+
+    def setup_method(self):
+        pass
+
+    def teardown_method(self):
+        pass
+
+    def test_urls(self):
+        client = Client()
+        assert(client.get('/').status_code == 200)
+        assert(client.get('/healthz/').status_code == 200)
+        assert(client.get('/runs/').status_code == 200)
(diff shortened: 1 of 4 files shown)


lightvector:
I'm going to go ahead to merge this.

lightvector:
Since this is just tests and some bugfixes.

src/frontend/utils/parse_util.py
```diff
@@ -0,0 +1,10 @@
+from django.http import Http404
+
+def parse_int_or_404(x):
+    if x:
+        try:
+            parsed = int(x)
+            return parsed
+        except ValueError:
+            raise Http404("Invalid argument")
```
tychota, line 9: Is 404 the right error for this. I would have said 400.

src/frontend/views/games_view.py
```diff
@@ -37,6 +38,8 @@ class GamesListByNetworkView(ListView):
   def get_queryset(self):
     self.run = get_object_or_404(Run, name=self.kwargs["run"])
     self.network = get_object_or_404(Network, name=self.kwargs["network"])
+    if self.network.run != self.run:
+      raise Http404("Network exists but is from the wrong run")
```
tychota, line 42: Is 404 the right error for this. I would have said 400, or 409.

src/frontend/tests/test_basic_urls.py
```diff
@@ -0,0 +1,195 @@
...
+from src.apps.games.models import RatingGame, TrainingGame, GameCountByNetwork, GameCountByUser
+from src.apps.games.tests.test_game import goodnpzbase64
+
+pytestmark = pytest.mark.django_db
+
+User = get_user_model()
+
+fake_sha256 = "12341234abcdabcd56785678abcdabcd12341234abcdabcd56785678abcdabcd"
+
+class TestUrlsEmptySite:
```
tychota, line 19: Nice

src/frontend/utils/parse_util.py
lightvector: Isn't 404 more consistent with what happens when in general you access a page that doesn't exist? Like katagotraining.org/networks/network-that-does-not-exist/?

src/frontend/views/games_view.py
lightvector: Same thing - you access a path that has the run, and then a network that doesn't exist for that run - that's a 404 right?

PR #77, opened 2021-01-18
Announcements

lightvector:
This adds a simple mechanism to add different sections to the front page with arbitrary html, from the admin page. Also it adds an acknowledgments line that we will use in the footer for KataGo's public run.


lightvector:
Also added a test.

src/apps/announcements/apps.py
```diff
@@ -0,0 +1,17 @@
+from django.apps import AppConfig
+from django.utils.translation import gettext_lazy as _
+
+
+class AnnouncementsConfig(AppConfig):
+    """
+    Katago Announcments app handles some messages for the front page of the site
```
tychota, line 7: Nits: Typo "Announc**e**m**e**nts"

src/apps/announcements/admin/announcement_admin.py
```diff
@@ -0,0 +1,36 @@
...
+
+class AnnouncementAdmin(admin.ModelAdmin):
+    """
+    AnnouncementAdmin allows admin to create or edit Announcements
+    """
+
+    list_display = (
+        "id",
+        "title",
+        "display_order",
```
tychota, line 13: nice the display order.

src/apps/announcements/admin/announcement_admin.py
```diff
@@ -0,0 +1,36 @@
...
+        "enabled",
```
tychota, line 14: And nice the enabled, would have done exactly the same.
:)

src/apps/announcements/models/announcement.py
```diff
@@ -0,0 +1,58 @@
...
+    title = CharField(
+        _("title"),
+        null=False,
+        blank=False,
+        max_length=240,
+        help_text=_("Title of the announcement section."),
+        db_index=True,
+        unique=True,
+    )
+    contents = TextField(
```
tychota, line 35: Did you prefer raw html as something like markdown ?

Easier to code, but maybe a bit less clear than md to write.
Plus security breach vector if at some points people other than me or you can write annoncement (stored XSS).

src/apps/announcements/models/announcement.py
```diff
@@ -0,0 +1,58 @@
...
+    id = BigAutoField(primary_key=True)
+    created_at = DateTimeField(_("creation date"), auto_now_add=True, db_index=True)
+    updated_at = DateTimeField(_("updated date"), auto_now=True, db_index=True)
+    title = CharField(
+        _("title"),
+        null=False,
+        blank=False,
+        max_length=240,
+        help_text=_("Title of the announcement section."),
+        db_index=True,
```
tychota, line 32: Do you really need an index here ?

tychota (approved):

src/apps/announcements/tests/test_announcements.py
```diff
@@ -0,0 +1,47 @@
+import pytest
+import base64
+
+from django.test import Client
+from src.apps.announcements.models import Announcement
+
+pytestmark = pytest.mark.django_db
+
+class TestAnnouncements:
```
tychota, line 9: nice

tychota:
Also do we aims to translate the website at some point (eg make a chineese version) ? 
Everything should be nice and easy to translate (thanks django po/mo system and {trans} balises) but the new announcement doesn't include anything to say "this blog post is in english and shall appaer only on english pages".

src/apps/announcements/models/announcement.py
lightvector: Thanks, made it markdown. But I still only intend admins to write. :)

src/apps/announcements/models/announcement.py
lightvector: Unique probably implies index practically, but I removed it to just go with the default, thanks.

src/apps/announcements/apps.py
lightvector: fixed

lightvector:
Going ahead to merge this, but I would be interested to hear your thoughts for how to produce a translated version. Thanks!

PR #81, opened 2021-01-23
Precommit hooks try #2

sanderland:


lightvector:
Seems this is all working without trouble. Thanks for the cleanup!

PR #85, opened 2021-01-26
Format number of games output

Nihisil:
These changes will make run stats output a bit more readable.

I didn't found `%d training games` strings in locale files, so I think it is safe to just change them here, and it will not affect other languages.


tychota (approved):
Looks fine for me but waiting for @lightvector to review before merging, as it uses to be a perf issue in this area and he solved this one.

@Nihisil did you test locally that page render with this change or should we (lightvector or I) do a local test

Nihisil:
Yep, I tested it, but only on simple data that I created manually. It looks like recreating all models and records manually will take a while. If you have a copy of production DB on your local installation it would be good if you can check how it looks as well. Thank you.

lightvector:
This seems great, thanks. I will test on my dev copy of the server and merge this in a day or two.

PR #86, opened 2021-01-26
Remove useless filter from admin that takes a really large space and cost extra SQL queries

tychota:


lightvector:
Seems good to me.

PR #87, opened 2021-01-26
Filter new net from elo graph

tychota:


lightvector:
Why not do this in the javascript? In the javascript we could also choose later to alter the drawing of the network in other ways instead of not drawing it at all, right?

tychota:
I found it easier to add it here, as the display logic is already long and hard to grasp on javascript side.

But yes this is more flexible to do it in JS.
My recommendation will still be to do it in django until we need to alter the drawing.
But if you want I can close this PR and do it in JS now.

tychota:
Also note that the choice made in django is really (maybe too) concervative.

image

So we will still have big bar, but a bit more precise than when there is zero rating games

lightvector:
ok sure let's go with that :)

PR #90, opened 2021-01-31
Add support for disabling games

lightvector:
Disabled games will not be used for rating. As for training, we'll handle that for now by a separate mechanism that still makes use of this field but queries the database manually outside the server since that's simpler at the moment.


lightvector:
Going ahead to merge since I think we want to use this right away.

PR #92, opened 2021-02-15
Fix rounding of elo_uncertainty

kcwu:


src/apps/trainings/models/network.py
```diff
@@ -215,7 +215,7 @@ class Meta:
     )
 
     def __str__(self):
-        return f"{self.name} ({self.elo}±{2 * self.elo_uncertainty})"
+        return f"{self.name} ({self.elo}±{2 * self.elo_uncertainty:.1})"
```
tychota, line 218: can you extract the ".1" to a variable like precision and replace 
`self.elo_uncertainty:.1` by `self.elo_uncertainty:.{precision}f`

that way we can reuse this precision for line 226 and 234

tychota (approved):
Apart from a small sugestion (that you are free not to accept ^^), looks amazing. I'm mind blown.

src/apps/trainings/models/network.py
kcwu: Done. However, I feel it might be not easier to read.

tychota (approved):

PR #93, opened 2021-03-02
Highlight strongest network in networks page

kcwu:
I wrote this code by imitating the code around it. I didn't set up the server and test it.

This is the intended output (this screenshot is created by modifying the style via browser dev console, not by real code testing).
image

### DIFF
--- a/src/frontend/static/css/custom.css
+++ b/src/frontend/static/css/custom.css
@@ -12,6 +12,6 @@
   font-weight:bold;
 }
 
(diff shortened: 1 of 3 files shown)


lightvector:
Cool.
Do you think the same coloring would be good to apply to the strongest confidently rated network message at the top of the table, to make the correspondence more obvious?

kcwu:
How about this?
image

lightvector:
Site is updated with this change. Looks good, thanks!

